PR #175, opened 2014-12-28
Allow going backwards in playlist

ids1024:
The makes the Q key go to the previous track. This branch includes the changes in PR #161, so that should probably be merged first.


np1:
Thanks, merged into https://github.com/np1/mps-youtube/tree/prev-track for now. Will merge into develop soon. I will also update the keymap help text. Any reason for using Q instead of a lowercase key?

ids1024:
You could use any key, but I though of Q since q is the key for going forward.

PR #176, opened 2015-01-06
Display volume when using mpv 0.7+ with json IPC

ids1024:
This makes the status bar display the volume with mpv 0.7+. Note that it doesn't show the volume until it is changed.


np1:
Good job, thanks. Hadn't noticed that this was missing. On my system it shows the volume whether I change it or not.

PR #180, opened 2015-01-09
Fix input_file for mplayer on Windows

ids1024:
Apparently mplayer cannot handle paths starting with a drive letter.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -1984,8 +1984,9 @@ def launch_player(song, songdata, cmd):
         if "mplayer" in Config.PLAYER.get:
             cmd.append('-input')
             if mswin:
-                # Mplayer does not recognize path starting with drive letter
-                input_file = input_file[2:]
+                # Mplayer does not recognize path starting with drive letter,
+                # or with backslashes as a delimiter.
+                input_file = input_file[2:].replace('\\', '/')
             cmd.append('conf='+input_file)
             p = subprocess.Popen(cmd, shell=False, stdout=subprocess.PIPE,
                                  stderr=subprocess.STDOUT, bufsize=1)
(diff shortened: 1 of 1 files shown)


ids1024:
Actually, I don't seem to have fixed it. I will keep working on this.

ids1024:
Now it should work.

np1:
Doesn't seem to work on Windows XP for me, it doesn't implement the key mapping (mplayer runs with default keys). I will try it in Windows 8 when I next reboot my machine.

ids1024:
I tested it on windows 7. Could it be because the temporary file path on XP contains spaces? I don't think that should be a problem, but it is a difference that comes to mind.

np1:
I don't think so, mpsyt passes the following to mplayer:
`-input conf=/docume~1/np1/locals~1/temp/mpsyt-inputhdjxuy`

When I try running mplayer directly on a file with this parameter it works and respects the key mapping. Just not when it is called from mpsyt.

I will investigate some more.

ids1024:
That would be the problem then. I'm not familiar with that formatting for a path with the tildes, but it does not seem to be used on Windows 7. Mplayer seems to expect paths to be formatted like Unix, so it cannot handle that dosism.

np1:
But mplayer does handle it when I enter it directly at the Windows command prompt;

eg, I enter:
`mplayer -input conf=/docume~1/np1/locals~1/temp/mpsyt-inputhdjxuy audio.m4a`

The file plays and then when I press `n` or `p` it exits as expected.

However, in mpsyt when I press `n` nothing happens and when I press `p` it pauses - so it is not using the key mapping. I will try to find out why this is.

ids1024:
That's strange. I don't know what the cause is then.

PR #181, opened 2015-01-10
Always pass exename to get_mpv_version

ids1024:
Trying mpsyt on Windows, I had trouble with this. Mpsyt would crash when it ran mpv.exe, so the player needed to be set to mpv.com. With this change and "set player mpv.com," it works.

This is also necessary if mpv is not in PATH.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -218,7 +218,7 @@ def get_config_dir():
     return mps_confdir
 
 
-def get_mpv_version(exename="mpv"):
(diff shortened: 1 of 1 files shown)


ids1024:
It also seems mpsyt crashes with a full path.  Known_player_set() returns True, but the player path does not exist in playerargs_defaults.

ids1024:
I think my second commit fixes the problem I mentioned. It now handles full paths and .com extensions fine.

np1:
Thanks again!

PR #184, opened 2015-01-11
Use return code to determine if player succeeded

ids1024:
When the player exited too quickly, for example skipping a couple songs or going back a couple, mpsyt retried. This was rather annoying.

I don't know if there was any reason for the previous method.


np1:
I believe the previous method was due to some players returning a non-zero return code in some situations even when the item played ok. I'm happy to run with this and add cases for other return codes if/when we notice them. Thanks :)

ids1024:
I don't know about mplayer, but mpv has an "Exit Codes" section in the man page. It says that 0 is used for success (as is the standard) and 1-3 are used for errors.

PR #185, opened 2015-01-11
Replace spaces with hyphens in 'open' and 'view' commands

ids1024:
It makes sense for open and view to behave the same way as save in this regard.


np1:
Thanks :+1: 
It would be good to get rid of the ugly hyphens completely at some point and handle spaces in playlist names using shlex. This would require a slight change in syntax from a user perspective, the user would need to quote playlist names when using some commands.

PR #187, opened 2015-01-13
Dead code removal / Minor bug fixes / Minor improvement

ids1024:


ids1024:
@np1 What purpose does this code serve? Mpsyt seems to run without it.

    __main__.Playlist = Playlist
    __main__.Video = Video

np1:
Good work, thanks a lot for the cleanup :)

I can recall that without those two lines the pickle module was not able unpickle the saved playlist data, Feel free to play around with it!

ids1024:
If so there should probably be a comment in the code, because it is not obvious at all.

np1:
I just tried removing the two lines and was unable to recreate the pickle error. Maybe they can be removed then. I tested with Python 2 and 3 running mpsyt directly and running after having installing with setup.py

Issue #178, opened 2015-01-08
how to disable input.conf override ?


spvkgn:
mps-yt from develop branch launches mpv with ```--input-conf=/tmp/mpsyt-input``` followed by override my input.conf

Is there a way not to override it?

Thanks.

np1:
Hi, thanks for testing the develop branch. I could add a `--no-input-override` flag to mpsyt to stop it doing this. It would then be advisable for those with a custom input.conf to map the following additional keys in their custom config;

`quit 42` for previous track
`quit` for next track
`quit 43` to exit a playing playlist

I could add an explanation of this in the wiki / faq

ids1024:
@np1 Do you think there would be any problems with reading the user's input.conf and replacing "playlist_prev" (pt_step -1 with mplayer) with "quit 42" and "playlist_next" ("pt_step 1" with mplayer) with "quit 43" (this would mean a minor change in how the exit codes are interpreted).

This way either standard mpv/mplayer keybinds would be used, or the user's custom ones.

np1:
I think it's a good idea. We would need to hunt around for the input.conf file (/etc/mplayer,  /usr/local/etc/mplayer, ~/.mplayer) wouldn't we?

ids1024:
Yes, it would require finding all the paths. Do you know where the file is stored on Windows? I also don't know if Mac is different for normal Unix.

ids1024:
Looking at mpv's code, this is what I get, in order:

- Specified at command line
- os.environ.get('MPV_HOME')
Windows (and not Cygwin)
------------------------
- os.environ['APPDATA']
- directory of mpv.exe
- directory of mpv.exe/mpv

Unix (including OSX and Cygwin)
-------------------------------
- os.environ.get('XDG_CONFIG_HOME', os.environ['HOME'] + '/.config') + '/mpv'
- os.environ.get('HOME') + '/.mpv' #Compatability with older versions
- if OSX: https://github.com/mpv-player/mpv/blob/master/osdep/path-macosx.m
- MPV_CONFDIR, which seems to be defined during compilation (could be problematic)

The manpage lists these directories, but the precedence is less clear. I haven't found the code specifying system level configuration files, but according to the manpage that would be at /usr/local/etc/mpv/, depending on --prefix during compilation.

Would it be better to just have mpsyt specific input.conf files (one for mplayer and one for mpv) that people can edit? Perhaps with the same replacement I mentioned, so that it can be just a symlink.

[...]

PR #188, opened 2015-01-14
Add mpv_input_conf and mplayer_input_conf settings

ids1024:
This is my solution to issue #178. With this, the user can have a custom input.conf.  The commands for going to the next and previous track are replaced with the quit commands that communicate that to mpsyt.

In addition, the defaults are changed to be the same as mpv/mplayer.  If users want to go through their playlist with j/k, they can set mpv_input_conf/mplayer_input_conf to an input.conf that does that.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -706,6 +706,8 @@ class Config(object):
     COLOURS = ConfigItem("colours",
                          False if mswin and not has_colorama else True,
                          check_fn=check_colours)
+    MPV_INPUT_CONF = ConfigItem("mpv_input_conf", '')
+    MPLAYER_INPUT_CONF = ConfigItem("mplayer_input_conf", '')
 
 
 class Playlist(object):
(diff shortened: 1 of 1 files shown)


mps_youtube/main.py
```diff
@@ -3130,11 +3150,11 @@ def play_range(songlist, shuffle=False, repeat=False, override=False):
         if returncode == 42:
             n -= 1
 
-        elif returncode == 43:
-            break
+        elif returncode in (43, None):
```
np1: Should this be (43, 0) ?

ids1024: No, playsong returns None for several errors. However, I notice that this will not properly handle returncodes 1-3, which should be interpreted as an order to go to the next song.

mps_youtube/main.py
```diff
@@ -3130,7 +3150,7 @@ def play_range(songlist, shuffle=False, repeat=False, override=False):
         if returncode == 42:
             n -= 1
 
-        elif returncode == 43:
+        elif returncode == 0:
             break
```
np1: If we break on returncode of 0 only the first item of a selection will be played.

ids1024: I didn't think of that.

np1:
HI, still not working right for me. With mpv it is fine. Have you tested with mplayer? Here `q` goes to next track (instead of exiting) and `>` doesn't do anything when I use mplayer. Not sure why at the moment. I have not set any values in the [mplayer/mpv]_input_conf setting.

ids1024:
It works for me with mplayer, but there seem to be issues if I exit mplayer too quickly.  In that case, it goes to the next track for <, >, and q.  I don't know why my issue or yours is occurring.

np1:
My bad, I had an invalid setting in `playerargs` left over from other testing, ignore my last comment, mplayer is fine. Sorry!

ids1024:
I don't know the cause of the problem I mentioned, but it seems to be caused by mplayer and it is not a major issue.  Mplayer is returning zero rather than the requested value for some reason.

Based on my attempts to find the cause of this problem, I pushed two small unrelated commits that I think help.  Also, I believe catching OSError is considered a bad practice, because it is a very general exception.  It might be better to catch a more specific error in some parts of the code.

np1:
Yep, I noticed that skipping between songs is more responsive with mpv than with mplayer, but as you mention, it's not a major issue - and thanks to your work it's a lot better than it was.

Issue #121, opened 2014-10-09
mps-youtub​e with portable Python or standalone on Windows


ZaxonXP:
Dear Nagev,

I found the mps-youtube tool very useful and I would like to use it in a portable mode.

Currently it requires the Python installation which is not so convenient. I tried to use it with the Portable Python but so far I failed to install the mps-youtube with it. The best would be to create the Windows exe file and put it on GitHub, but I am not so advanced in Python yet to achieve this.

Could you please add some instructions on the GitHub page how to install it on the Portable Python or better provide precompiled version for Windows?

Best regards,
ZaxonXP

ids1024:
A standalone binary can now be built with `python setup.py py2exe`, assuming py2exe is installed. A prebuild binary will be available with the next release.

ids1024:
Available with latest release. See the download attached to the release:

https://github.com/np1/mps-youtube/releases/tag/v0.2.5

ids1024:
@ZaxonXP Have you tried the standalone binary provided with the latest release? I don't ordinarily run Windows and would like to know if it is working properly.

PR #190, opened 2015-01-16
Support py2exe

ids1024:
This is a solution to Issue #121.  It seems to work, but will probably need more testing.  Run `python setup.py py2exe` to create a "dist" folder with mpsyt.exe and it's dependencies.

Edit: This requires py2exe of course.  Just run `pip install py2exe` to install that.


np1:
Good job, thanks again, I will test this when I get a chance.

np1:
Did some basic testing on Windows 8 and it worked well :+1:

PR #195, opened 2015-01-22
Use default useragent

ids1024:
According to the git log, this code is from pms (now mps).  I presume this is because the service mps uses has a useragent test, but this is unnecessary with YouTube.


np1:
Thanks

Issue #132, opened 2014-10-28
Add MultiConnection Download


xubuntu4iran:

ids1024:
What about a `download_command` option, so that a user could make mpsyt download a video using programs like aria2c?

ids1024:
@np1 You can probably mark this as "implemented" now.

np1:
Version 0.2.2 released with `download_command` config option

type `help dl-command` for help.

To enable 4 connections for downloads using aria2c:
```
set download_command aria2c -k 1M -s4 -x4 --dir=%d --out=%f %u
```

PR #198, opened 2015-01-23
Add download_command option

ids1024:
This is a solution to issue #132.

To make mpsyt use aria2c to download:

    set download_command aria2c -d / -o '%f' '%u'


np1:
Thanks, I tested this and it seems to work well.
I got aria2c to use four connections with: 

`set download_command aria2c -s4 -x4 -d / -o '%f' '%u'`

Note that the download file needs to be sufficiently large for aria2c to use multiple connections (20MB per connection), otherwise use the `-k` flag to set a lower minimum split size.

This implementation could be improved with more substitutions, e.g,

%d - download directory (the value set in mpsyt DDIR config setting)
%F - complete path of download (currently implemented as %f in this PR)
%f - download filename (`os.path.basename` of above)

This would be more natural and allow for more flexibility. It also prevents the user having to specify the directory as `/`.

PR #203, opened 2015-01-26
Make tuple_xy default to None rather than (None, None)

ids1024:
This is an alternative to PR #200.  That pull request incorrectly changes the test for the return value of `_get_terminal_size_windows()`.  This also makes the return values of `_get_terminal_size_windows()` and `_get_terminal_size_linux()` consistent, which makes sense.


np1:
Thanks guys :+1:

PR #204, opened 2015-01-27
Fix race condition in mpv 0.7+ code

ids1024:
I have noticed that occasionally the progress bar would not show up.  My interpretation is that mpv sometimes takes longer that the one second wait to set up the socket, and thus an error was raised, exiting player_status.  The player would keep playing, due to the `p.wait()` statement.  If I am correct, this should solve the problem.

I do not know if there is a better way to do this.


np1:
Thank you for the fix. I haven't used mpv 0.7 much so haven't seen this issue.

PR #210, opened 2015-02-06
Mpris

### LINKED ISSUE #191: Mpris interface
Implementing mpris would give us capability to control  mpsyt remotely. Plus mpris is freedesktop standard, so lot of things would just work on most systems with mpsyt. I mean cool things like media keys support or integration to desktop (yes, it would allow mps to be controlled from GNOME default applet).

Cons: it would bring `dbus` dependency and it's not working on windows by design (thus it would broke nothing for windows).

I'm willing to implement this, if it would be accepted.

hrnr:
This pull request implements MPRIS2 remote interface.

This would close #191.


np1:
Thanks for all the effort put into this @hrnr & @ids1024 

Can you change the license to GPLv3?

Do you know which exception you are checking for [here](https://github.com/np1/mps-youtube/pull/210/files#diff-6345c7136a08fb23636add591c756ed0R906), if so can you specify it?

I am getting this message when I run mpsyt:
```
                                                   _         _
       _ __ ___  _ __  ___       _   _  ___  _   _| |_ _   _| |__   ___
      | '_ ` _ \| '_ \/ __|_____| | | |/ _ \| | | | __| | | | '_ \ / _ \
      | | | | | | |_) \__ \_____| |_| | (_) | |_| | |_| |_| | |_) |  __/
      |_| |_| |_| .__/|___/      \__, |\___/ \__,_|\__|\__,_|_.__/ \___|
                |_|              |___/ v0.2.1








Enter /search-term to search or [h]elp
ERROR:dbus.service:Unable to append ({'HasTrackList': False, 'CanQuit': False, 'Fullscreen': False, 'CanRaise': False, 'SupportedMimeTypes': [], 'CanSetFullscreen': False, 'Identity': 'mps-youtube.instance11256', 'SupportedUriSchemes': []},) to message with signature a{sv}: <type 'exceptions.ValueError'>: Unable to guess signature from an empty list
> 
```

It subsequently runs fine although I'm not sure whether mpris is working as I get no response when trying to use the media keys during playback.

hrnr:
I fixed all issues reported by landscape, including all exceps.  d2a27f1 should fix the error you spotted on startup, although I can't reproduce. Since python-dbus is partially based on guessing, current approach should be more safe.

hrnr:
I'm not sure about licence, if it should be GPLv3, is only a small notice (_"this is part of project and sharing licence"_), like in `setup_py2exe.py` enough? Or should there be a full licence?

Media keys should work out-of-the-box for people using `gnome-settings-daemon` since this [bugfix](https://bugzilla.gnome.org/show_bug.cgi?id=697810). KDE's now playing widget should support mpris2 as well. I'm not personally  using any desktop enviroment, just the awesome WM, so a can't test it right now, but I plan to download some recent images and test the default config.

ids1024:
See the license notice at the top of `main.py`.  It is the standard notice for files using the GPL.

hrnr:
ok, I will just copy that.

hrnr:
ok, I did some testing and push above 2 commits fixing minor issues. I tested mainly applets provided by DEs:

* fedora 21 (GNOME, mpv 0.7.3, [Media player indicator](https://extensions.gnome.org/extension/55/media-player-indicator/)) - works great
* kubuntu 14.04 (KDE, mpv 0.3.4, "Now playing" plasmoid) - works
* ubuntu 14.04 (that-ubuntu's-DE, mpv 0.3.4, default music applet) should support mpris, but not working
* mint 17.1 (cinnamon, mpv 0.3.4, default music applet) works

I would call that a pretty good support. I'm not sure what is the issue with ubuntu, but I would suspect whitelisting, since everywhere else it just works.

For media keys I couldn't get them forwarded to virtual machine, so I didn't test that, but I suppose they will be working in DEs where default applets are working.

np1:
Good job with all the testing.

I'm also not able to get it working with xfce in xubuntu 14.04 (tried with mplayer 1.1-4.8 and mpv 0.3.4). It doesn't appear in the default applet and doesn't respond to media keys.

hrnr:
Isn't the applet same in ubuntu and xubuntu? I will try to find the source code, and see if we can do anything about that.

hrnr:
I've gone throught their source code and found the problem. They are using desktop files everywhere.

I pushed some changes to [ubuntu](https://github.com/hrnr/mps-youtube/tree/ubuntu) branch, that should fix that, but it requires installed desktop file. I put some example desktop file to root.

I had big troubles to get it actually working, because of ubuntu's caches (even `desktop-file-install` was not working). So I don't know if this file is working, but it is valid desktop file and from code it doesn't seem that they are requiring any special parts. I succeded with `sudo mv /usr/share/applications/evince.desktop /usr/share/applications/mps-youtube.desktop`, which finally made applet work -- all works fine next/prev, play, pause and metadata. To cache problems restarting session should work, but I was working from live media in virtual machine, so I don't know (and I hate ubuntu all by now, so I don't want to install).

IF we want to support ubuntu we need to ship and install `.desktop` file. Or we could just add that 1 line in `mpris.py` and put some info in manual for ubuntu users to install it themselves. I'm not sure what would be the right way..

hrnr:
I finally solved cache problems and `.desktop` in ubuntu branch works fine for me with just 
```
sudo desktop-file-install mps-youtube.desktop
```
(it basically just copies the file to `/usr/share/applications`)

ids1024:
Is it ordinary for command line programs to provide .deskop files?  It doesn't appear so to me.  That said, it seems alright if it is required to get this working on Ubuntu.  A .desktop could be included with the code but not installed by `setup.py`.

hrnr:
I agree, so far I think that the best solution would be to add that 1 line referencening desktop file to `mpris.py`, and put somewhere a note how to get it working on ubuntu. 

For `.desktop` file there could be a gist, it doesn't need to be in this repo.

If it will be working for @np1 on xubuntu, I will add the line to `mpris.py` and it could be merged.

np1:
Hi, it's appearing in the applet for me in xubuntu using the Ubuntu branch and having run `sudo desktop-file-install mps-youtube.desktop`. It's not responding to any of the media keys though (play/pause, next, prev). I'll see if I can figure out why..

ids1024:
> For .desktop file there could be a gist, it doesn't need to be in this repo.

It can, but it would probably be easier just to put it somewhere in the repo.  There is really no reason not to include it.

ids1024:
@np1 Does the applet respond to media keys when using a player other than mpsyt?  It might just by that the applet you're using doesn't do that or is not configured to do so.

hrnr:
I'm not sure if ubuntu's media keys handler supports mpris, it surely didn't a year ago, as there are many [tutorials](http://www.teknically-speaking.com/2013/02/make-your-media-keys-work-with-spotify.html) how to make ubuntu's media keys work with spotify. Same things should now apply on mpsyt, it uses the same interface.

However applet's controls should work, it definitely uses mpris (if the applet is the same as in ubuntu). Also applet should show song title when playing.

nivit:
Hi, I just tested this version with Gnome Shell under FreeBSD, and all works fine: applet and media keys (pause, play, next/previous song):

mps_youtube-top-bar-system-settings

mps_youtube-top-bar-next-song

np1:
Yes, the applet is showing the title of the playing item correctly.

The media keys work with one player in the applet (Quod Libet) but not others (Clementine, VLC, Audacious), so I guess in Ubuntu the applet (sound indicator) doesn't respect the media keys.

Regarding installation of the .desktop file, we could create a python script to generate it and run the `sudo desktop-file-install` command, which could be triggered using a command such as `mpsyt --install-desktop-ubuntu`

I don't like that mps-youtube appears with a "no icon" graphic (the red X)

applet

It makes it look inferior! I haven't tried it with any other DE's to check how it appears [edit: looks good in the images posted by @nivit]

An icon could be created and specified in the desktop file, or perhaps a stock icon can be specified (such as the terminal icon).

ids1024:
> An icon could be created and specified in the desktop file, or perhaps a stock icon can be specified (such as the terminal icon).

How does the ascii art shown when opening mpsyt look as an icon in that size?  If it looks good use that, otherwise stick to a stock icon.

np1:
@ids1024 I will give it a go.

Another thought, since we can get the YouTube thumbnail jpg of playing items fairly easily, would it be easy to integrate this into the applet display so that it shows that in the same way the applet shows album cover art for other players?

ids1024:
The thumbnail would be good.  I looked at the API docs.  Defining `mpris:artUrl` in `Metadata` should do it.

ids1024:
We should probably define the `LoopStatus` and `Shuffle` properties, since mpsyt supports looping and shuffling and this would be easy to implement.  Just send these to the mpris process from in the `play_range()` function.

We could also possibly define more metadata (Album, Artist, etc.) when the `album` command is used.

hrnr:
Problem with `LoopStatus` and `Shuffle` is those are read/write properties, if implementing them, we should also implement setters.

Having nicer metadata would be nice, it would allow us to send some nice notifications (with art image and so on) .

For media keys, I'm really glad it's working in gnome (thanks @nivit). In ubuntu media keys are not handled by applet, it's handled directly by unity (support for media keys was dropped from `gnome-settings-daemon` which did it for most DEs, it's not true anymore as I found). For xfce it must be some xfce part. This is not a dealbreaker, media keys can be bind to [playerctl](https://github.com/acrisci/playerctl) or similar.

ids1024:
It isn't possible to implement those as read only?  I suppose if those could be set over mpris, they should be settable through keybindings in the player.  We can't really do that currently, but it might be possible later with a more elaborate interface.  We can skip it for now.

hrnr:
We could implement them as read-only. We could just ignore any Set calls on those properties, which we do by default. But once we implement them there is not way to inform clients that we are hadling them as read-only. Specs are just read/write or nothing.

Other POV is: would it broke some clients/applets badly? I don't think so. I'm not sure which clients shows this properties, nor how which sets them, but I think is just could broke the shuffle button.

ids1024:
Let's not.  It may not break much, but it probably does little good either.

np1:
I have merged current develop branch with hrnr's ubuntu branch into [mpris branch](https://github.com/np1/mps-youtube/tree/mpris)

I added the YouTube thumbnails with these two commits;
https://github.com/np1/mps-youtube/commit/e3620a08d2f1ecb248bec5de5b6d15cc01c5612f
https://github.com/np1/mps-youtube/commit/f211a3d2e3a93885050de049b91d80ae3ce0b687

It's working ok for me, tested only on Xubuntu 14.04

screenshot - 170215 - 13 53 21

I would prefer if sound-indicator didn't try to squeeze the textual metadata into one line, other applications have it split by title/artist/album. 

I still have to write the code to generate and install the ubuntu desktop file.

ids1024:
The indicator in kde plasma 5 is working well (including the thumbnail).  I installed kde to test this, since I normally just use dwm.

hrnr:
Thumbnails are great. It would be nice if thumbnails would be also available for notifications too, but that's maybe feature for different PR/branch.

As far as I know getting title/artist/album info is diffucult. It would be possible with [youplay](https://github.com/apassant/youplay), which is based on [freebase](https://developers.google.com/freebase/) and heuristics. I was getting impressive results from that. Problem is that freebase is deprecated and will be down soon. Google should present some transition options during March. Then hopefully yp could be ported.

I was experimenting with splitting video title by "-" + simple heuristics, but it was very unreliable.

hotice:
Why is the text in the Ubuntu Sound Menu greyed out for mps-youtube?

PR #221, opened 2015-02-25
Install desktop file on first run on Ubuntu systems

np1:
This installs the `mps-youtube.desktop` file needed to get the MPRIS interface working in Ubuntu / Xubuntu sound indicator. It gets installed automatically when running mpsyt and put in `~/.local/share/applications` so doesn't need sudo.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -852,6 +852,7 @@ def init():
     init_readline()
     init_cache()
     init_transcode()
+    init_ubuntu_desktop_file()
 
     # set player to mpv or mplayer if found, otherwise unset
     suffix = ".exe" if mswin else ""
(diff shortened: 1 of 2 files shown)


ids1024:
Is this really better than just installing it in `setup.py` on all Linux systems? It seems like a bit of a hack to add it at run time and only on Ubuntu.

np1:
AFAIK we only need it on Ubuntu / Xubuntu to get MPRIS interaction working so I don't see the point of adding it to all systems.
Yes it is a hack, if you have a more elegant solution installing it with setup.py I would be happy to merge it.

ids1024:
Apparently you can add `NoDisplay=true` to the desktop file to make it not display in menus.  I think it would be best to do that and then install on all systems.  It is probably not needed except on Ubuntu, but there would be no harm in it's presence.  There may possibly be other software that does the same thing as Ubuntu's indicator.

PR #222, opened 2015-02-27
Debian directory

zgrimshell:
This is debian/ directory needed for proper building of Debian packages (aka .deb files). Notice also that commits to this directory are or will be the one found in official Debian packages as I am the current maintainer there.


np1:
Thanks I have merged this into [Debian branch](https://github.com/np1/mps-youtube/tree/Debian)

PR #225, opened 2015-03-04
Added Dockerfile and running mpsyt in a container instructions

rothgar:
I'm already using the container myself (it's really basic) but thought I'd push it up stream in case anyone else wants to use it too.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:latest
+MAINTAINER Justin Garrison <justinleegarrison@gmail.com>
+
+RUN apt-get update && apt-get install -y mplayer
+RUN pip install mps-youtube
+RUN apt-get clean && apt-get purge
+
+ENTRYPOINT ["mpsyt"]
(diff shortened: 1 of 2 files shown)


ids1024:
I've never used docker, but it looks like it could be a useful way to install mps-youtube, so I guess this can be merged.

rothgar:
@leothelocust if you want video to work you'll need to pass through the display socket and set your host X session to allow it to be used.

On your host you should run `xhost +` and when you run the container you need to add the following options
```
-v /tmp/.X11-unix:/tmp/.X11-unix -e DISPLAY=unix$DISPLAY
```

I'm not 100% sure that's all you'll need but that usually works for basic applications.

rothgar:
I just confirmed this works and I pushed a new version of the container.

selection_020

I used 

```
sudo docker run -it --rm -v /tmp/.X11-unix:/tmp/.X11-unix -e DISPLAY=unix$DISPLAY -v /dev/snd:/dev/snd --name mpsyt rothgar/mpsyt
```

leothelocust:
Thanks @rothgar!  I'm trying it out right now.

Issue #226, opened 2015-03-20
Set title for the window


r8b7xy:
It'd be useful to show current playing track in window title.
e.g. 'urxvt - The Rolling Stones - Start Me Up'

PR #229, opened 2015-03-20
Set window title

ids1024:
This implements #226.  It should work under windows, but I haven't tested it.


np1:
Good job, I haven't tested on Windows either. Let's hope it works.

PR #236, opened 2015-04-02
Fix errors about trying to access methods of NoneType / sockpath not found

ids1024:


ids1024:
I added a possible fix for #243. It may not solve the root problem there, but it should probably have this test anyway since mpv may not have created the socket yet.

PR #247, opened 2015-04-20
Use LINES and COLUMNS for termsize of dumb terminals

punchagan:
Using `fcntl.ioctl` returns (0, 0) termsize on a dumb Emacs terminal,
but 'LINES' and 'COLUMNS' environment variables give the correct size.


np1:
Thanks :)

Issue #254, opened 2015-05-06
Search don’t display results


Kaliiixx:
When I search a video, one result appear : 
 1   https://youtube.com/devicesupport                                                        03:56
But playlists research displays well.

guaraqe:
It is related to this issue with this pafy, which is used by mps-youtube, and it is due to an API change by YouTube.
https://github.com/np1/pafy/issues/78
Also:
https://github.com/np1/mps-youtube/issues/249

dabn:
Oh, this is a sad day.
So, guess I am going back to watching youtube on youtube.com with double the cpu workload again. My battery life will be crippled.

Kaliiixx:
I think I mispoke.
This result is the uniq one, regardless  of the query

RobertoGongora:
This is due to the API being deprecated :(

ids1024:
I have created a branch with a very incomplete search functionality based on gdata 3:
https://github.com/ids1024/mps-youtube/tree/gdata3wip

Currently it only can display page one of results and all videos appear to have a length of 20 sec. It is probably a bad idea to try to add a video to a playlist on this branch, because it will save the wrong metadata. Caveat emptor.

lol768:
Nice work, @ids1024. I know it's not grabbing all the data yet but it's neat how few changes to the existing codebase were required to get it to show a page of results. It seems pretty snappy too.

It unfortunately [looks like](http://stackoverflow.com/a/18138843) you can't request the "contentDetails" part along with a search request, so there's no way to get information about the duration etc without at least [one](https://developers.google.com/youtube/v3/docs/videos/list#try-it) additional HTTP requests. Maybe YouTube could add this.

I believe I'm correct in stating you can only get the following info for each video via the search endpoint, which doesn't seem like too much:

```json
{
   "snippet":{
      "publishedAt":"2014-02-12T21:01:24.000Z",
      "channelId":"UCJ6td3C9QlPO9O_J5dF4ZzA",
      "title":"[Happy Hardcore] - nanobii - Rainbow Road [Monstercat Release]",
      "description":"Support on Beatport: http://monster.cat/1gqSVQY Support on Bandcamp: http://monster.cat/1eUyTym Support on iTunes: http://monster.cat/1g2TwX9 --- Listen on ...",
      "thumbnails":{
         "default":{
            "url":"https://i.ytimg.com/vi/a0Aauep0VWs/default.jpg"
         },
         "medium":{
[...]

Issue #249, opened 2015-04-21
https://youtube.com/devicesupport on search results


spvkgn:
I have noticed it today. On first num 'https://youtube.com/devicesupport' from 'youtubehelp' user appears with every search or user list.

knutshub:
I'm not completely sure. But I think this is the link to the API version 3 page including the relevant information for searching:
https://developers.google.com/youtube/v3/docs/search

spvkgn:
I don't fully understand why this link appears. Would be nice find a way to remove it from search results.

GijsTimmers:
Related: https://github.com/rg3/youtube-dl/issues/5483

JamesTheAwesomeDude:
@GijsTimmers I don't think it's just related; I think it's a full duplicate.

The other issue is also about the exact same thing: ytsearch only showing that error page.

ids1024:
See np1/pafy#78

Mpsyt may soon stop working without a fix. That video is not the main issue.

RPiAwesomeness:
Yeah, @ids1024 is right. If you go to http://www.youtube.com/devicesupport there's a page about the Youtube API being upgraded and "Certain older Youtube apps" not working after late May.

@np1 Gotta fix this D:

JamesTheAwesomeDude:
rg3/youtube-dl#5483 was resolved with rg3/youtube-dl@b4c0806

Does that resolve this as well?

lol768:
The rg3/youtube-dl solution seems preferable to me than using the newer API.

JamesTheAwesomeDude:
It's a little funny to think that, as @lol768 pointed out, a third-party borderline TOS-violating ripper written in Python is actually going to end up being a more reliable and useful API than anything provided by Google themselves..

I guess that's just the world we live in these days. :expressionless:

jrenner:
I saw this as top search result for a couple weeks. Now it is the only result I see returned from any search,.

lol768:
@jrenner Yeah, this is how they approached/are approaching phasing out the API. They first added the result, and are now replacing it entirely.

> In early May, Data API v2 video calls will start returning only the warning video introduced on April 20. Users will not be able to view other videos on apps that use the v2 API video calls. See youtube.com/devicesupport for affected devices.

[...]

PR #256, opened 2015-05-08
Use the gdata 3 API (WIP)

ids1024:
This is a work in progress update to use gdata 3, so that mpsyt will actually work. Comment here if you have fixed something in your own repo and I will  merge it into the branch. I think that is a little better than just commenting on a github issue and exchanging commit urls.

#249 #254 
@np1 @jkatzwinkel @lol768

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -2464,16 +2464,21 @@ def usersearch(q_user, page=None, identify='forUsername', splash=True):
         # if the user is looked for by their display name,
         # we have to sent an additional request to find their
         # channel id
-        url = "https://www.googleapis.com/youtube/v3/channels"
-        query = {'part':'contentDetails',
-                 identify: user,
-                 'key': Config.API_KEY.get}
+        url = "https://www.googleapis.com/youtube/v3/search"
+        query = {'part': 'id,snippet',
+                 'maxResults': 1,
+                 'q': user,
+                 'key': Config.API_KEY.get,
+                 'maxResults': 2,
+                 'type': 'channel'}
 
         try:
             userinfo = json.loads(utf8_decode(urlopen(url+
                 "?"+urlencode(query)).read()))['items']
             if len(userinfo) > 0:
-                channel_id = userinfo[0].get('id')
+                snippet = userinfo[0].get('snippet')
+                channel_id = snippet.get('channelId')
+                user = snippet.get('title')
             else:
                 g.message = "User {} not found.".format(c.y + user + c.w)
                 return
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -785,8 +785,8 @@ class g(object):
     user_cache = collections.OrderedDict()
     model = Playlist(name="model")
     last_search_query = {}
-    current_pagetoken = None
-    page_tokens = [None]
+    current_pagetoken = ''
+    page_tokens = ['']
     active = Playlist(name="active")
     text = {}
     userpl = {}
@@ -1527,23 +1527,23 @@ def get_tracks_from_json(jsons):
 
     # delete global page token list if apparently new search
     if not g.current_pagetoken:
-        g.page_tokens = [None]
+        g.page_tokens = ['']
 
     # make page token list from api response
     page_tokens = [jsons.get(field) for field in
-        ['prevPageToken', 'nextPageToken']]
+                   ['prevPageToken', 'nextPageToken']]
     page_tokens.insert(1, g.current_pagetoken)
 
     curidx = lambda: g.page_tokens.index(g.current_pagetoken)
 
     # update global page token list
     if page_tokens[2]:
-        if page_tokens[0]:
+        if page_tokens[0] and g.page_tokens[curidx()-1] is None:
             g.page_tokens[curidx()-1:curidx()+2] = page_tokens
         else:
             g.page_tokens[curidx():curidx()+2] = page_tokens[1:]
     else:
-        if page_tokens[0]:
+        if page_tokens[0] and g.page_tokens[curidx()-1] is None:
             g.page_tokens[curidx()-1:curidx()+1] = page_tokens[:2]
         else:
             g.page_tokens[curidx()] = page_tokens[1]
@@ -1558,7 +1558,7 @@ def get_tracks_from_json(jsons):
     vurl += "?" + urlencode({'part':'contentDetails,statistics,snippet',
                              'key': Config.API_KEY.get,
                              'id': ','.join([get_track_id_from_json(i)
-                                             for i in items]) })
+                                             for i in items])})
     try:
         wdata = utf8_decode(urlopen(vurl).read())
         wdata = json.loads(wdata)
@@ -2448,7 +2448,7 @@ def generate_search_qs(term, page=None, result_count=None, match='term'):
     if page:
         qs['pageToken'] = page
     else:
-        g.current_pagetoken = None
+        g.current_pagetoken = ''
 
     if Config.SEARCH_MUSIC.get:
         qs['videoCategoryId'] = 10
@@ -2544,12 +2544,12 @@ def usersearch_id(q_user, page=None, splash=True):
         g.message = msg
         g.last_opened = ""
         g.last_search_query = {"user": q_user}
-        g.current_pagetoken = page
+        g.current_pagetoken = page or ''
         g.content = generate_songlist_display(frmat="search")
 
     else:
         g.message = failmsg
-        g.current_pagetoken = None
+        g.current_pagetoken = ''
         g.last_search_query = {}
         g.content = logo(c.r)
 
@@ -2571,13 +2571,13 @@ def related_search(vitem, page=None, splash=True):
         g.message = "Videos related to %s%s%s" % (c.y, ttitle, c.w)
         g.last_opened = ""
         g.last_search_query = {"related": vitem}
-        g.current_pagetoken = page
+        g.current_pagetoken = page or ''
         g.content = generate_songlist_display(frmat="search")
 
     else:
         g.message = "Related to %s%s%s not found" % (c.y, vitem.ytid, c.w)
         g.content = logo(c.r)
-        g.current_pagetoken = None
+        g.current_pagetoken = ''
         g.last_search_query = {}
 
 
@@ -2598,13 +2598,13 @@ def search(term, page=None, splash=True):
         g.last_opened = ""
         g.last_search_query = {"term": term}
         g.browse_mode = "normal"
-        g.current_pagetoken = page
+        g.current_pagetoken = page or ''
         g.content = generate_songlist_display(frmat="search")
 
     else:
         g.message = "Found nothing for %s%s%s" % (c.y, term, c.w)
         g.content = logo(c.r)
-        g.current_pagetoken = None
+        g.current_pagetoken = ''
         g.last_search_query = {}
 
 
@@ -4020,7 +4020,7 @@ def user_more(num):
         g.content = generate_songlist_display()
         return
 
-    g.current_pagetoken = None
+    g.current_pagetoken = ''
     item = g.model.songs[int(num) - 1]
     user = g.meta.get(item.ytid, {}).get('uploader')
     usersearch(user, identify='id')
@@ -4034,7 +4034,7 @@ def related(num):
         g.content = generate_songlist_display()
         return
 
-    g.current_pagetoken = None
+    g.current_pagetoken = ''
     item = g.model.songs[int(num) - 1]
     related_search(item)
 
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -784,9 +784,8 @@ class g(object):
     url_memo = collections.OrderedDict()
     model = Playlist(name="model")
     last_search_query = {}
-    current_page = None
-    next_page = None
-    prev_page = None
+    current_pagetoken = None
+    page_tokens = [None]
     active = Playlist(name="active")
     text = {}
     userpl = {}
@@ -1519,13 +1518,34 @@ def get_track_id_from_json(item):
     return ''
 
 
-def get_tracks_from_json(jsons):
-    """ Get search results from web page. """
 
-    page_tokens = {'next_page': 'nextPageToken',
-                   'prev_page': 'prevPageToken'}
-    for var, field in page_tokens.items():
-        setattr(g, var, jsons.get(field))
+
+
+def get_tracks_from_json(jsons):
+    """ Get search results from API response """
+
+    # delete global page token list if apparently new search
+    if not g.current_pagetoken:
+      g.page_tokens = [None]
+
+    # make page token list from api response
+    page_tokens = [jsons.get(field) for field in
+        ['prevPageToken', 'nextPageToken']]
+    page_tokens.insert(1, g.current_pagetoken)
+
+    curidx = lambda: g.page_tokens.index(g.current_pagetoken)
+
+    # update global page token list
+    if page_tokens[2]:
+        if page_tokens[0]:
+            g.page_tokens[curidx()-1:curidx()+2] = page_tokens
+        else:
+            g.page_tokens[curidx():curidx()+2] = page_tokens[1:]
+    else:
+        if page_tokens[0]:
+            g.page_tokens[curidx()-1:curidx()+1] = page_tokens[:2]
+        else:
+            g.page_tokens[curidx()] = page_tokens[1]
 
     items = jsons.get("items")
     if not items:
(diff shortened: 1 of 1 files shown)


JKatzwinkel:
I think I fixed the `user <username>` and `user <username>/<query>` commands. Also two minor bugs in songlist display generation which are my own fault.

ids1024:
I had noticed `user` was not working but didn't get around to fixing it. Now it seems to work after pulling your changes.

lol768:
Thoughts on https://github.com/lol768/mps-youtube/commit/76165720b739686ff94d88dc75fef8973f143a62?

ids1024:
@lol768 It looks reasonable. I don't think it would be good to require people to provide an api key, but having it as an option with a valid key as the default seems like a good compromise. Ultimately it depends on what @np1 thinks of it, but for now I've merged it into the wip branch.

lol768:
https://github.com/lol768/mps-youtube/commit/bdbfc98569d91b37a226203119e5f948778f956b fixes https://landscape.io/github/ids1024/mps-youtube/19/modules/mps_youtube/main.py#L2449

lol768:
Since you worked on this part, @JKatzwinkel:

Right now the usersearch works great (nice to be able to listen to music in the background again) but it does feel a little slow, no doubt as a result of the number of requests you need to perform to get useful info from the API. When I timed it, it look 2.3s to grab a page of results. I feel we could do one or more of the following to speed it up:

* Retrieve the playlist id by sending a request to the Channel list endpoint and requesting only the `id` field: https://www.googleapis.com/youtube/v3/channels?part=id&forUsername=mrsuicidesheep&fields=items%2Fid

This returns a response similar to the following:
```json
{
 "items": [
  {
   "id": "UC5nc_ZtjKW1htCVZVRxlQAQ"
  }
 ]
}
```

You can work out the playlist ID of the user's uploads by taking the channel id and replacing the second character with a 'U'. This gives "U**U**5nc_ZtjKW1htCVZVRxlQAQ" and reduces the quota cost of every channel lookup (because we don't need to request contentDetails).

* Cache usernames and their associated upload playlist ID for a set amount of time. Unless a user changes their username, the upload playlist ID (should?) stay the same.
* Cache video IDs and durations for an (infinite?) period of time. As far as I know, durations of videos can never change so it'd be safe to hold this data forever and if all videos in a page were cached in this way it'd mean one less request.

Do these sound reasonable/workable? I'd be happy to have a go at adding in some or all of these.

JKatzwinkel:
@lol768 I already noted your suggestion of requesting only specific reponsefields yesterday and wanted to be on it, but then somehow forgot about it. I also thought about caching some of those split up information in a central store similar to how it is already being done with search requests and am glad to see its not only me. What really impresses me and will probably make things alot easier is the channelId <->playlist id hack you suggest there. How did you even come up with that? I'm looking forward to trying it myself.
I think all those ideas are great. If you feel like splitting up, I could also happily help implementing your suggestions, but I won't get to do any work before tomorrow I think.

ids1024:
I have noticed a problem with user searches (and probably other searches, but I haven't seen it yet). Some videos are region locked, and thus won't play. For example, some results for `user rushvevo` are allowed only in Canada, and thus won't work for my IP in the US. `contentDetails.regionRestriction` provides this information, but I can't seem to find a simple and reliable way to get the country code of the user's IP. Does anyone know a good way to do that or another way to filter out region locked videos?

lol768:
> I already noted your suggestion of requesting only specific reponsefields yesterday and wanted to be on it, but then somehow forgot about it. I also thought about caching some of those split up information in a central store similar to how it is already being done with search requests and am glad to see its not only me.

Sorry if I've said the same suggestion multiple times, my memory isn't too great - it's nice to have everything in one PR though now so I can look back over stuff.

Yeah, I'd forgotten there was caching code already in the codebase. Haven't looked too much at how it works but hopefully we can reuse that logic.

> What really impresses me and will probably make things alot easier is the channelId <->playlist id hack you suggest there. How did you even come up with that? I'm looking forward to trying it myself.

So I'd noticed that the channel IDs and user upload playlist IDs looked similar. At first I thought they were both identical, but the second letter always seemed to differ. I tried it on a few channels and it seemed to work:

* Proximity: UC3ifTl5zKiCAhHIBQYcaTeg => UU3ifTl5zKiCAhHIBQYcaTeg
* MonsterCat: UCJ6td3C9QlPO9O_J5dF4ZzA => UUJ6td3C9QlPO9O_J5dF4ZzA
* MrSuicideSheep: UC5nc_ZtjKW1htCVZVRxlQAQ => UU5nc_ZtjKW1htCVZVRxlQAQ
* TastyNetwork: UC0n9yiP-AD2DpuuYCDwlNxQ => UU0n9yiP-AD2DpuuYCDwlNxQ

If I had to speculate, I'd assume the C stands for Channel and the U for uploads. My only worry is YouTube changing this scheme in the future but it seems to work well right now.

> I think all those ideas are great. If you feel like splitting up, I could also happily help implementing your suggestions, but I won't get to do any work before tomorrow I think.

Sure, I'd be happy to look at a implementing some of those suggestions tomorrow. If you'd like, I can focus on implementing the hack above and tweaking the requests so they only get the data we need (at a lower quota cost) along with fixing some of the style issues Landscape has flagged up if you're happy to look at some of the caching stuff?

---
> I have noticed a problem with user searches (and probably other searches, but I haven't seen it yet). Some videos are region locked, and thus won't play. For example, some results for user rushvevo are allowed only in Canada, and thus won't work for my IP in the US. contentDetails.regionRestriction provides this information, but I can't seem to find a simple and reliable way to get the country code of the user's IP. Does anyone know a good way to do that or another way to filter out region locked videos?

Could you perform a WHOIS on the user's IP using the ARIN WHOIS database and look at the country returned from that? E.g.:

```
➜  ~  whois 216.58.208.78

#
# ARIN WHOIS data and services are subject to the Terms of Use
# available at: https://www.arin.net/whois_tou.html
#
# If you see inaccuracies in the results, please report at
# http://www.arin.net/public/whoisinaccuracy/index.xhtml
#


#
# The following results may also be obtained via:
# http://whois.arin.net/rest/nets;q=216.58.208.78?showDetails=true&showARIN=false&showNonArinTopLevelNet=false&ext=netref2
#

NetRange:       216.58.192.0 - 216.58.223.255
CIDR:           216.58.192.0/19
NetName:        GOOGLE
NetHandle:      NET-216-58-192-0-1
Parent:         NET216 (NET-216-0-0-0-0)
NetType:        Direct Allocation
OriginAS:       AS15169
Organization:   Google Inc. (GOGL)
RegDate:        2012-01-27
Updated:        2012-01-27
Ref:            http://whois.arin.net/rest/net/NET-216-58-192-0-1



OrgName:        Google Inc.
OrgId:          GOGL
Address:        1600 Amphitheatre Parkway
City:           Mountain View
StateProv:      CA
PostalCode:     94043
Country:        ***US***
RegDate:        2000-03-30
Updated:        2013-08-07
Ref:            http://whois.arin.net/rest/org/GOGL


OrgAbuseHandle: ZG39-ARIN
OrgAbuseName:   Google Inc
OrgAbusePhone:  +1-650-253-0000 
OrgAbuseEmail:  arin-contact@google.com
OrgAbuseRef:    http://whois.arin.net/rest/poc/ZG39-ARIN

OrgTechHandle: ZG39-ARIN
OrgTechName:   Google Inc
OrgTechPhone:  +1-650-253-0000 
OrgTechEmail:  arin-contact@google.com
OrgTechRef:    http://whois.arin.net/rest/poc/ZG39-ARIN


#
# ARIN WHOIS data and services are subject to the Terms of Use
# available at: https://www.arin.net/whois_tou.html
#
# If you see inaccuracies in the results, please report at
# http://www.arin.net/public/whoisinaccuracy/index.xhtml
#

➜  ~  
```

ids1024:
The arin.net whois api should work, but is there a way to query information about the IP accessing the site? Otherwise another request would have to be made to a site like whatismyip.org to get the user's external IP address.

JKatzwinkel:
@lol768 However you like. I will see what I can do caching-wise. 
@ids1024 Even though I am in germany where practically everything is geoblocked, I haven't noticed this yet. I suspect the reason is that with the new API, youtube video search does not even list items with your location blacklisted. How annoying is that?
Determining the running mpsyt instance's countrycode with `whois`  seems a great idea, but yes, finding out a host's gateway public IP address is not too easy.

ids1024:
@JKatzwinkel As I said above, the issue was with the `user` command. The youtube search api may already filter those out. If you try `user rushvevo` and select a geoblocked video like `Superconducter` you should get an error.

ids1024:
I found a solution to the problem. I noticed that `user rushvevo/` should only the accessible videos, so I changed it to use the same query. This also greatly simplifies that function.

JKatzwinkel:
I can see that, nicely done! To be honest though, I thought of this behaviour more as a feature than a problem. Listing all results and giving a good error msg for geoblocked tracks at least tells you what there actually is and youtube web interface searches also used to work that way. until recently. But this is not the place to discuss that maybe I can elaborate sometime after the api transition issues are fixed.

lol768:
@ids1024 Interesting that blocked videos show in the user's uploads but not in the search. I like that we can reuse the existing search code now, I hadn't realized you could search by channel like that using the search endpoint. It also removes the need for grabbing the playlist ID which is neat.

ids1024:
@JKatzwinkel The thing is that it wasn't giving a good error message, just `Youtube says: This video is not available.` That also seems to be all the player on youtube.com gives, and the videos are not shown on youtube.com either. I don't see why you would want to see those videos listed. It is just annoying trying results and see that a third of them are unavailable why I try to play them.

If it is a genuinely useful feature, there should be a config option for it, but we should worry about that after everything else is working.

JKatzwinkel:
@ids1024 I agree.

lol768:
@ids1024 

Using 4f54405 @ gdata3wip (https://github.com/ids1024/mps-youtube/commit/4f54405d9e964676a564fb9fc6c387c40c6cca49) the user search appears to be broken.

**Steps to reproduce:**

1. Type "user monstercatmedia" and hit return.
1. Observe that the status message doesn't make sense: "Results for  (by monstercatmedia)" (should be just "Video uploads by monstercatmedia").
1. Observe that only one video is shown in the list ("1 [Trance] - Au5 - Crossroad (feat. Danyka Nadeau) [Monstercat Release]") and that the video is incorrect (if order is set to date, the latest video should be shown, which has the title "[Trance] - Trivecta - Evaporate (feat. Aloma Steele) [Monstercat Release]").
1. Observe that https://www.youtube.com/user/MonstercatMedia/videos shows many more videos and a different set of videos when ordered by date.

ids1024:
I changed the message to `Videos uploaded by monstercatmedia`, as it should be. As for the other issue, that is because the other videos are not in the "music" category. Use `set search_music false` and they will be listed. Do you think non-music categories should not be filtered out on user search?

lol768:
> As for the other issue, that is because the other videos are not in the "music" category. Use set search_music false and they will be listed. Do you think non-music categories should not be filtered out on user search?

Ah, my apologies - I didn't realise they were categorised like that. It works perfectly when search_music is false - thanks for clarifying that. 

Yeah, I'm wondering if all videos should be shown for user search because right now if I set search_music to false it means I can't do a keyword search restricted to music. If not, I guess you could add another config option search_music_user or something, though that might be a bit confusing.

Odd that the MonsterCat people seem to categorise their videos as "Entertainment" instead of "Music".

lol768:
Right now user search pagination fails because of the added parameter to the *usersearch* function.

https://github.com/lol768/mps-youtube/commit/e92af1d8773e8f309875b3ea7b3ad376741bd345 is how I tried to fix it. It works but because of YouTube's new pagination, I can't think of an easy way to have "user <username> <pagenum>" working again (other than grabbing 1 to (n-1) results pages to work out the token for the requested page).

ids1024:
> I can't think of an easy way to have "user " working again (other than grabbing 1 to (n-1) results pages to work out the token for the requested page).

Isn't the only way to get to a page (other than the first one) by the next page command from the previous page? So, page tokens for previous pages could all be stored in a list, such as `g.page_tokens`. That way, `g.page_tokens[0]` would be page 1, etc. The next token would be added when a page that lies at the end of the list is opened. Do you think that would work/make sense/be a good idea?

JKatzwinkel:
This is what I had in mind. If you think this is the way to go, I'm on it this instant.

ids1024:
> This is what I had in mind. If you think this is the way to go, I'm on it this instant.

You mean what I suggested in the post immediately above yours? In that case, go for it.

lol768:
> Isn't the only way to get to a page (other than the first one) by the next page command from the previous page? So, page tokens for previous pages could all be stored in a list, such as g.page_tokens. That way, g.page_tokens[0] would be page 1, etc. The next token would be added when a page that lies at the end of the list is opened. Do you think that would work/make sense/be a good idea?

Sorry, GitHub interpreted my comment as HTML.

I originally meant to write:

> I can't think of an easy way to have "user \<username\> \<pagenum\>" working again (other than grabbing 1 to (n-1) results pages to work out the token for the requested page).

However, as far as I can see, the command never actually supported a pagenum argument so the point is moot.

JKatzwinkel:
I am not aware of a way to get the list of all page tokens involved in a result set from the API. But if the pagenum argument wasn't implemented until now, then probably nobody have missed it so far. 
Page token list is coming up though...

JKatzwinkel:
@lol768 It was a way bigger challenger than I expected, but browsing through pages somehow works not with my revision jkatzwinkel@a7db186a9d0fda9fda2accdf947fff936c93cc4e . I see that you have begun to work around the now page change mechanism for playlists with which the searches working with API v3 are now inconsistent. I have done something about this, too. Maybe you want to have a look at my changes before continuing to try yourself so we don't do it twice. It is not ready but as it took this long, I just wanted to have something pushed.

ids1024:
> Should we change playlist search to API 3.0 as well or just leave it like it is because the old gdata calls still work?

They probably won't work for long, so it would be best to change it now.

JKatzwinkel:
@ids1024 I changed an api call in usersearch to improve it and made a first attempt to have user searches cached to spare api request. Did not get to update playlist searches yet, so in case any of you guys wants to give it a shot, feel free to do so. Working on video category resolution for now.

JKatzwinkel:
@np1 @ids1024 @lol768 Pickle user account data cache, fix issues in pagination, calm down pylint a bit. Is it ok to save username search result cache (matching user name + channel Id) to the same file as streams cache? jkatzwinkel@22c4b227e9d62e200d588f2850a60bf876f35135

ids1024:
> Is it ok to save username search result cache (matching user name + channel Id) to the same file as streams cache?

That sounds fine to me.

JKatzwinkel:
@lol768 @ids1024 At this point, I think rewriting the remaining old API calls can safely be taken care off after this PR will been merged, given that they still work for some reason. To me it seems that no really pressuring tasks remain with the search implementations for now.

lol768:
@JKatzwinkel I've been impressed with it so far. Been using it daily with no complaints or errors to report. There are a few styling nitpicks that could be fixed but nothing major as far as I can see.

ids1024:
http://youtube-eng.blogspot.co.uk/2015/04/bye-bye-youtube-data-api-v2.html
> By late May, v2 API calls except for comments and captions will receive 410 Gone HTTP responses. You can test your application’s reaction to this response by pointing the application at eol.gdata.youtube.com instead of gdata.youtube.com. While you should migrate your app as soon as possible, these features will work in the Data API v2 until the end of July 2015 to avoid any outages.

It looks like it will stop working by the end of this month, so I think we stop using gdata 2 at all by the next release of pafy and mpsyt (which should be very soon, so that people can run mpsyt without installing from git).

I am working on porting the playlist search now.

JKatzwinkel:
Very well, I'm looking forward to taking a look in the morning. I'm afraid I have no idea of pafy whatsoever so far, but maybe I'll get around to read up on it. Still, with the video search seemingly working now, maybe an intermediate release of mpsyt would be in order so that people get reassured that there's progress? That way some extra time could be bought for porting the playlist seach.

ids1024:
It is working well currently, so perhaps it would be good to release. There would probably be another release not all too long after, but I don't see any problem with that and it is rather urgent since mpsyt is largely nonfunctional as of the last release.

@np1 What do you think of merging the current gdata 3 code and having a release?

ids1024:
@np1 I think it is time for a release. I see you have added me as a collaborator on github, but I would need access to pypi as well to do a release myself. If you don't have anything against releasing a new version of mpsyt now, would you either add me to pypi (I have the same username there) or do the release yourself?

np1:
@ids1024 I've added you on pypi now, thanks

PR #276, opened 2015-05-16
Display category names in songlists, user name in user upload search messages

JKatzwinkel:
These changes include two output prettifications that didn't make it into v0.2.4 release:
  * Lookup **video category names** mapped to category Ids for printing in songlist display. [77baf141df24b9b674c357eae325c2ecc6e63aeb, 3ab04c5e53c2a3fec8ee5217922b0a1b90bed8e4]
  * Unknown or stale categories are updated using the `videoCategories` API endpoint
  * Category data is saved to cache file
  * Show **user name** instead of channel Id **in status/result message** when fetching user uploads as invoked by `u <num>` command. [1c5cb629f282b9c72bff41e341c9a4366252a52d]

Additional changes that don't really do anything:
  * Cosmetic changes where made in order to resolve some issues landscape.io had with the code, in an attempt to get project health score up in new release.
  * I also changed position of the optional parameter `splash` in the signature of method `usersearch()` so it does match the order parameters are expected by all the other search methods. But that doesn't affect anything at all as far as I'm aware.


JKatzwinkel:
Bonus patch: fix video duration parsing error fd3662fd4dfa588258a7e5384e4dd6457e5f6989

ids1024:
It looks like pafy needs this functionality, based on the failed travis builds: https://travis-ci.org/np1/pafy

I guess this could be moved to pafy and accessed in mps-youtube where needed as well...

PR #281, opened 2015-05-17
fix landscape bot

hrnr:
Adds compatibility import to help landscape parse `main.py`. This import should do no harm with python 3, it just does nothing.


ids1024:
I guess there's no harm in adding back that code. I hope landscape add py3 support soon.

hrnr:
wow that was fast. Landscape with python 3 support would be best.

PR #284, opened 2015-05-17
Some fixes related with new YouTube playlist processing

JKatzwinkel:
Changes described in commit list below.


ids1024:
Pagination of playlists still needs to be fixed...

JKatzwinkel:
I had it fixed by using only calls to the new API instead of `pafy.get_playlist`, but that is probably not a good idea. (See c989ada11ce85944a9c0487f5f74d556ebe77551). However, I can fix pagination of pafy-powered playlist contents tomorrow, that's not a big deal afair.

PR #289, opened 2015-05-22
Reworked desktop file install

hrnr:
Current version doesn't remove desktop file on uninstall. Also checking in runtime seems to be a bit hack.

This version uses setup tools to install desktop file where appropriate (it will be installed system-wide, instead of user directory).

Tested on ubuntu 14.04 with `pip3 install https://github.com/hrnr/mps-youtube/archive/desktop-file.zip --upgrade`. MPRIS is immediately functional after that.

### DIFF
--- a/install/linux/usr/share/applications/mps-youtube.desktop
+++ b/install/linux/usr/share/applications/mps-youtube.desktop
@@ -1,0 +1,11 @@
+[Desktop Entry]
+Name=mps-youtube
+GenericName=Music Player
+Keywords=Audio;Song;Podcast;Playlist;youtube.com;
+Exec=mpsyt %U
+Terminal=true
+Icon=terminal
+Type=Application
+Categories=AudioVideo;Audio;Player;
+StartupNotify=true
+NoDisplay=true
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -873,7 +873,6 @@ def init():
     init_readline()
     init_cache()
     init_transcode()
-    init_desktop_file()
 
     # set player to mpv or mplayer if found, otherwise unset
     suffix = ".exe" if mswin else ""
@@ -1061,46 +1060,6 @@ def init_cache():
             dbg(c.r + "Cache file failed to open" + c.w)
 
         prune_streams()
-
-
-def init_desktop_file():
-    """ Install desktop file. """
-    # Required for mpris on Ubuntu
-    if mswin:
-        return
-    appdir = os.path.expanduser("~/.local/share/applications")
-    filename = "mps-youtube.desktop"
-
-    if not os.path.exists(os.path.join(appdir, filename)):
-        dbg("no desktop file in local appdir")
-
-        if has_exefile("desktop-file-install"):
-            dbg("has desktop-file-install executable")
-
-            if not os.path.exists(appdir):
-                os.makedirs(appdir)
-
-            fname = os.path.join(tempfile.gettempdir(), "mps-youtube.desktop")
-
-            desktop_file_contents = \
-                "[Desktop Entry]\n" \
-                "Name=mps-youtube\n" \
-                "GenericName=Music Player\n" \
-                "Keywords=Audio;Song;Podcast;Playlist;youtube.com;\n" \
-                "Exec=mpsyt %U\n" \
-                "Terminal=true\n" \
-                "Icon=terminal\n" \
-                "Type=Application\n" \
-                "Categories=AudioVideo;Audio;Player;\n" \
-                "StartupNotify=true\n" \
-                "NoDisplay=true"
-
-            with open(fname, "w") as dtf:
-                dtf.write(desktop_file_contents)
-
-            cmd = ["desktop-file-install", "--dir={}".format(appdir), fname]
-            subprocess.call(cmd)
-            os.unlink(fname)
 
 
 def init_readline():
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ python setup.py sdist bdist_wheel
 """
 
 from setuptools import setup
+import sys
 
 options = dict(
     name="mps-youtube",
@@ -54,6 +55,12 @@ options = dict(
                        "bundle_files": 1}},
     long_description=open("README.rst").read()
 )
+
+if sys.platform.startswith('linux'):
+    # Install desktop file. Required for mpris on Ubuntu
+    options['data_files'] = [
+        ('share/applications/',
+            ['install/linux/usr/share/applications/mps-youtube.desktop'])]
 
 try:
     import py2exe


hrnr:
directory structure as `install/linux/usr/share/applications/mps-youtube.desktop` as maybe a bit clumsy. I'm not sure where this file should belong, please move it if appropriate.

ids1024:
What if the package is installed as an egg (I don't know how python eggs work, but I don't think this would work)?

hrnr:
It uses standard setuptools features, so it should work. Even if it didn't I don't think it would be a big issue as it only breaks MPRIS integration on ubuntu.

What I see as a big issue is that we don't uninstall properly, leaving left-over desktop file, which currently affects *all* non-windows systems.

But I think it will just work by setuptools magic, I will test that.

ids1024:
Running your `setup.py`, I don't see any `mps-youtube.desktop` file in `/usr/share/applications`. I don't think what you are trying to do is really supported by setuptools, even if it is possible.

https://pythonhosted.org/setuptools/setuptools.html#non-package-data-files
> The distutils normally install general “data files” to a platform-specific location (e.g. /usr/share). This feature intended to be used for things like documentation, example configuration files, and the like. setuptools does not install these data files in a separate location, however. They are bundled inside the egg file or directory, alongside the Python modules and packages.
> ...
> Note, by the way, that this encapsulation of data files means that you can’t actually install data files to some arbitrary location on a user’s machine; this is a feature, not a bug. You can always include a script in your distribution that extracts and copies your the documentation or data files to a user-specified location, at their discretion.

I think the current method, though imperfect, is the only way supported by setuptools, although I would love to be proven wrong. Is it really an issue if the desktop file is not removed on uninstall? It does not do anything (the `NoDisplay` line makes it not show up in menus).

hrnr:
Ok you were right, it doesn't work with `easy_install` althought desktop file is copied to egg properly. But I don't think it's an issue since `easy_install` is in fact deprecated and [not recomended](https://packaging.python.org/en/latest/installing.html#use-pip-for-installing) anymore.

With pip it works great. IMHO it's [supported](https://docs.python.org/3.4/distutils/setupscript.html#installing-additional-files) feature.

Depending on your setup, desktop file doesn't need to be in `/usr/share/applications`. I think it will more likely be in `/usr/local/share/applications`. Or just try `find / -name 'mps-youtube.desktop' 2> /dev/null`.

BTW regarding distribution we should not have `setup.cfg`, or configure package as non-universal, as it now only supports python3 as mentioned here http://pythonwheels.com/

ids1024:
> With pip it works great. IMHO it's supported feature.

That's the distutils documentation. Distutils supports it, but setuptools does not. I wasn't using easy_install, I just ran `python setup.py install`. I don't know why it is working for you in pip.

Good point with `setup.cfg`. I have pushed a commit deleting it.

hrnr:
Then it should be 
```
python3 setup.py install
python3 setup.py install_data
```

I'm just thought that we only have setup tools now.

ids1024:
Ah, I see now. I didn't know data files were installed with a separate command. I suppose then that this works, but not with an egg. Also, the setuptools documentation explicitly says that it is not supported, even though it seems to work. I guess this is probably better, but I still feel a little unsure about the fact it doesn't work with eggs.

ids1024:
The file also does not appear to be included in a wheel, though it is probably possible to fix that.

hrnr:
Desktop file not being in wheel would be a real problem.

hrnr:
If I create wheel with `python3 setup.py bdist_wheel` I have desktop file under `mps_youtube-0.2.4.data/data/share/applications`. I don't know if that is correct or not.

ids1024:
Here is the contents of a wheel created with `python setup.py bdist_wheel` on my system:

    $ unzip -l dist/mps_youtube-0.2.4-py3-none-any.whl
    Archive:  dist/mps_youtube-0.2.4-py3-none-any.whl
      Length      Date    Time    Name
    ---------  ---------- -----   ----
           30  2015-05-20 20:32   mps_youtube/__init__.py
        17347  2015-05-20 20:32   mps_youtube/mpris.py
         2867  2015-05-20 20:32   mps_youtube/terminalsize.py
       162078  2015-05-22 16:34   mps_youtube/main.py
         3977  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/DESCRIPTION.rst
           49  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/entry_points.txt
         1747  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/metadata.json
           12  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/top_level.txt
           92  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/WHEEL
         5548  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/METADATA
          940  2015-05-22 17:00   mps_youtube-0.2.4.dist-info/RECORD
    ---------                     -------
       194687                     11 files

If I then try to install the wheel with `sudo pip install -U dist/mps_youtube-0.2.4-py3-none-any.whl`, no desktop file is installed to `/usr/share/applications`.

ids1024:
Oops, looks like I was on the wrong branch. It seems to work.

hrnr:
And here is mine
```
$ unzip -l dist/mps_youtube-0.2.4-py2.py3-none-any.whl
Archive:  dist/mps_youtube-0.2.4-py2.py3-none-any.whl
  Length      Date    Time    Name
---------  ---------- -----   ----
      237  2015-05-22 18:05   mps_youtube-0.2.4.data/data/share/applications/mps-youtube.desktop
   160712  2015-05-22 18:44   mps_youtube/main.py
    17347  2015-05-16 12:51   mps_youtube/mpris.py
       30  2015-05-16 12:51   mps_youtube/__init__.py
     2867  2015-05-17 15:21   mps_youtube/terminalsize.py
     3977  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/DESCRIPTION.rst
       49  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/entry_points.txt
     1747  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/metadata.json
       12  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/top_level.txt
      110  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/WHEEL
     5548  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/METADATA
     1064  2015-05-23 02:06   mps_youtube-0.2.4.dist-info/RECORD
---------                     -------
   193700                     12 files
```

ids1024:
Ok, looks fine I guess. It is broken on `easy_install`, but if I recall correctly `easy_install` is deprecated in favor of `pip` anyway. It is not vital to the functionality of the program anyway, and is needed (as far as we can tell) only on Ubuntu.

PR #301, opened 2015-06-01
Updated debian directory

zgrimshell:
Now building and using properly python3

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -7,6 +7,7 @@ mps-youtube (0.2.5-1) unstable; urgency=medium
   * mps-youtube is now python3 only so build changes are made
   * Updated to latest upstream for Youtube APIv3 (Closes:#785704)
   * Small manpage tweak
+  * debian/rules: using now pybuild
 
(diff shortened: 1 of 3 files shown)


hrnr:
Hi, have you thought about updating debian package? Because debian package in repo is very old and it is not working since apiv2 is down.

zgrimshell:
@hrnr yes, new version of pafy and mps-youtube should hit unstable in next 24 hours if some problem doesn't pop out. The migration process to testing is 10 days if there is no problem. For Debian Jessie (the current stable) I will need to wait next round of jessie-updates (probably for 8.2 as 8.1 windows is closed) or to prepare backports (I'll investigate options). Also with debian dirs in place you can build packages yourself (and they will be the same as the official ones).

hrnr:
Great news! I'm glad mps-youtube is not dead on debian. I've made jessie packages with pbuilder and they work like a charm. thanks

zgrimshell:
Not dead at all. The packages were always updated in debian mentors and targeting debian experimental but they were not uploaded to it during the Jessie freeze and now I just merged all commits into one. Glad to help and good to know that it works.

PR #310, opened 2015-06-09
Fixed linux installation bug (py2exe related)

stephenbalaban:
Also cleaned up tab/whitespace issues in setup.py

Setup.py is now pep-8 compliant.

Previously, this error was being thrown after importing py2exe:

File "/usr/local/lib/python3.4/dist-packages/py2exe/_wapi.py", line 4, in
<module>

      _kernel32 = WinDLL("kernel32")

      NameError: name 'WinDLL' is not defined

This happened on my Ubuntu 14.04 LTS system.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,10 +48,13 @@ options = dict(
         "Development Status :: 5 - Production/Stable",
         "License :: OSI Approved :: GNU General Public License v3 (GPLv3)"
     ],
-    options={"py2exe": {"excludes": "readline, win32api, win32con, xerox, "
-	                            "dbus, gi, urllib.unquote_plus, "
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -65,7 +68,8 @@
     # in easy_install
     options['console'] = ['mpsyt']
     options['zipfile'] = None
-except ImportError:
+except (ImportError, NameError):
```
stephenbalaban: This is the actual fix. It simply catches the NameError that occurs.

ids1024:
Thanks for finding this issue. You probably shouldn't have py2exe installed on you Linux system though.

PR #316, opened 2015-06-18
WIP Plugin API Code

### LINKED ISSUE #314: Plugin API
It's great to see the code now split up into more modular pieces across multiple files. To take this further, it'd be interesting to see some sort of API which allowed users to load their own scripts which extend the core. This could perhaps also reduce the required maintenance of core functionality.

Ideally users could place a script file in a folder somewhere and have them loaded on startup. Scripts/plugins could potentially:

* Register their own commands
* Add new help entries
* Listen for common events (e.g. a new song has been played, a user performed a search, an item was added to a playlist)

Does this sound like it'd be useful to have? It's inspired to an extent by weechat's script system.

ids1024:
Initial Implementation of #314.

I created this pull request to help organize development of this feature.

As of this post, the feature does not work. The album search plugin can be loaded with `pluginload Album Search`, but mpsyt currently crashes when the `album` command is used.

This will resolve #314.


lol768:
> That does seem like a reasonable way to interact with other plugins. But then, the plugins may need access to the config option provided by core mpsyt functionality, such as search_music.

Access to core settings is probably going to be the most likely usecase. I can't think of many situations where a plugin would need access to the config from another unless one was an extension of the other/they had overlapping functionality.

> And then there is the question of how to access them. The list of config items does not seem like a particularly good way for the plugin to access them, since it would have to iterate over them and check the name. Would it help to add a method to Plugin to do that iteration, or is there a better way to define config options?

Perhaps if each plugin had a unique identifier, we use a dictionary: `{"lastfm_scrobble": {"api_key": ConfigItem(...)}}` This would mean much more efficient looking up of individual config items and would also enable easy retrieval of all configuration items for a plugin. This would presumably also make cleaning up stored settings on plugin uninstallation much easier. We could use some sort of notation, e.g. `lastfm_scrobble.api_key` and provide helper functions to process this notation and lookup the supplied item.

ids1024:
Now, should we also namespace the options from the core mpsyt functionality? They would then be something like `mpsyt.order`, `core.order`, or `mps-youtube.order`. I am not sure which is best. That would, among other things, eliminate the potential issue of options added to mpsyt conflicting with plugin names.

Perhaps plugins could specify their options in a list as specified earlier, but there could be a `ThePlugin.config` to access config options, allowing access only to the plugin's own configuration, and global configuration. This would by implemented by a classmethod property of `Plugin`. It would allow things like this: `ThePlugin.config.theplugin.option1` and `ThePlugin.config.core.option1`.

ids1024:
We need a higher level interface for `g.content` `g.message` `g.current_page`, etc. I think it would make sense to use a `Page` class.

Page would have methods `Page.next()` and `Page.prev()` to get the next and previous `Page` object, or `None` if there is none. It may be good to have a way to determine if there is a next or previous page without retrieving it, with something like `Page.has_next()` and `page.has_prev()`.

There should also be a `Page.num` to store the page number, and perhaps `Page.total_pages` for a count of how many pages there are.

For common functionality, like a list of songs (currently handled by `generate_songlist_display()`), a subclass of `Page` could be used perhaps.

lol768:
> We need a higher level interface for g.content g.message g.current_page, etc. I think it would make sense to use a Page class.

> Page would have methods Page.next() and Page.prev() to get the next and previous Page object, or None if there is none. It may be good to have a way to determine if there is a next or previous page without retrieving it, with something like Page.has_next() and page.has_prev().

> There should also be a Page.num to store the page number, and perhaps Page.total_pages for a count of how many pages there are.

Is it always possible to know how many pages are available?

> For common functionality, like a list of songs (currently handled by generate_songlist_display()), a subclass of Page could be used perhaps.

I think this makes sense. A consistent way of displaying error messages would also be nice too.

-----

One of the things I was wondering about whilst looking at the current code was the command regexes. What happens in a situation where more than one regular expressions match the user input? Do we somehow favour one plugin over another? Just refuse to do anything? I don't think calling them both would be especially helpful. Also, some plugins might want to override functionality/commands present in another plugin.

ids1024:
If the number of pages is not available, that could just return `None`.

The current behavior for multiple regexs matching is to just use whichever is first in `g.commands`. That way, built in commands would have precedence, then plugins, in the order the plugins are loaded. This is not by design, but simply the way the simple iteration code ended up working.

Calling both could be useful, but that should probably only happen when explicitly specified. Perhaps there could be a syntax like `@command(regex, ..., behavior='override'` to override a build in command, or some other value (`'redundant'`? I can't think of a better name right now) to run both. Otherwise it could just produce an error.

If we just silently ran both regexs, you could unintentionally end up two plugins defining the same command, which would have undesirable and strange results.

ids1024:
So, how should we handle plugins that depend on non-standard libraries? For example, the clipboard functionality. The plugin could just display a message about the library being missing when a command is run (current behaviour), but it seems like it would be better to handle that a different way.

lol768:
> So, how should we handle plugins that depend on non-standard libraries? For example, the clipboard functionality. The plugin could just display a message about the library being missing when a command is run (current behaviour), but it seems like it would be better to handle that a different way.

Perhaps we could have a list of modules that the plugin depends on, and refuse to load it if they are not all present?

ids1024:
Well, obviously we have to load the module before knowing what it depends on (to access the `Plugin` subclass). So I guess the plugin should probably import the modules it needs in `__init__`. Instead of having a list of what it needs, we can just catch an import error when creating an instance of `Plugin`.

It doesn't seem perfect, but I don't know what else we can do.

lol768:
> Well, obviously we have to load the module before knowing what it depends on (to access the Plugin subclass). So I guess the plugin should probably import the modules it needs in __init__. Instead of having a list of what it needs, we can just catch an import error when creating an instance of Plugin.

> It doesn't seem perfect, but I don't know what else we can do.

Ah yeah, that's a good point. I'm not particularly fond of doing the import in the init function.

We could potentially extract the metadata into another file (maybe JSON) and that way we'd be able to parse it and look at dependencies first. Having a separate metadata file would probably also be advantageous for other services that want to make use of the plugin information - e.g. a site where you can download plugins would be able to display versions, dependencies etc without having to execute code first. It'd also separate out the metadata and actual plugin functionality. It might also simplify plugin loading - the metadata file could specify the name of the module including the main plugin class and then we'd just need to scan the plugins folder for *.json files.

ids1024:
Hm, that could work. Although I liked how clean and simple the `loadPlugins()` function is, and it will get much more complicated if we are only loading some of the plugin's modules. We could also avoid loading the plugin modules until the plugin is actually activated.

So, do we just have a .json file in the same directory as the plugin, with the same name? So there is `theplugin.json` and `theplugin.py` (or the directory `theplugin`) next to each other in the `plugins` folder. Or, perhaps a better way would be if we put them together in a directory, or perhaps better, a zip file. Python has support for loading modules from a zip file, so that could work well.

ids1024:
There is an issue with the current plugin loading code for plugins included with mpsyt that is probably even worse with the method above-it does not work with the py2exe standalone windows binaries of mpsyt. We probably want the included plugins built into the binary, but that could be complicated to get working.

lol768:
This seems to work:

```python
import json
import re
from zipfile import ZipFile
import zipimport


def load_plugin(name: str):
    with ZipFile(name, 'r') as zip:
        items = zip.namelist()
        matches = [m.group(1) for l in items for m in [re.search(r'^(.*\.json)$', l)] if m]
        metadata = json.loads(zip.open(matches[0], 'r').read().decode("utf-8"))
        print(metadata)
        module = zipimport.zipimporter(name).load_module(metadata['name'])
        cls = getattr(module, metadata['name'])
        cls()

if __name__ == "__main__":
    load_plugin("MyPlugin.zip")
```

In the ZIP:

```json
{"name": "MyPlugin", "author": "lol768", "version": "1.5", "module_dependencies": ["zipfile"], "plugin_dependencies": ["some_other_plugin"]}
```

```python
class MyPlugin:

    def __init__(self):
        print("Hello from plugin!")
```

ids1024:
Yes, that works. We can make the `__init__` function of `Plugin` store the json metadata in the object (e.g. `theplugin.name`) so that it can be made retrievable from the plugin or other plugins without duplicating the metadata.

But what do we do about the plugins that are included with mpsyt? Do we install them as zips somewhere as well? That does not seem to be an option for the py2exe standalone binary. If we do make them zips, we would need those to be generated by `setup.py` somehow, because we do not want to store them as zips in the git repo.

lol768:
> But what do we do about the plugins that are included with mpsyt? Do we install them as zips somewhere as well? That does not seem to be an option for the py2exe standalone binary. If we do make them zips, we would need those to be generated by setup.py somehow, because we do not want to store them as zips in the git repo.

We could download them at install-time. if we had a GitHub organisation, we could have a repo per script (or one repo with branches in) - GitHub will then happily make us a ZIP archive based on the current code.

ids1024:
> We could download them at install-time. if we had a GitHub organisation, we could have a repo per script (or one repo with branches in) - GitHub will then happily make us a ZIP archive based on the current code.

That could work.

@np1 What do you think about moving mps-youtube (and probably pafy too) to a github org?

But then, we would need some way to actually develop the plugins. Perhaps we could allow either folders or zip files? That way, the git repo (or a symlink to it) could be placed in the plugins directory and function the same way as it does with a zip.

Or maybe we would be better off abandoning the zip idea and just using directories, and then including them with the rest of the code. I still can't think of a way to make that work with py2exe, however. Even if we do this, we should probably have it support zips, since that would be a more convenient way to distribute third party plugins.

lol768:
I have no real preference for folders vs ZIP files. I guess the nice thing about the ZIPs is that it's just a single file to distribute but it doesn't seem like it'd be too much work to support both:

Quick prototype:
```python
import json
import re
from zipfile import ZipFile
import zipimport
import os
from importlib import import_module

def load_plugin_zip(name: str):
    with ZipFile(name, 'r') as zip:
        items = zip.namelist()
        matches = [m.group(1) for l in items for m in [re.search(r'^(.*\.json)$', l)] if m]
        metadata = json.loads(zip.open(matches[0], 'r').read().decode("utf-8"))
        print(metadata)
        module = zipimport.zipimporter(name).load_module(metadata['name'])
        cls = getattr(module, metadata['name'])
        cls()

def load_plugin_directory(name: str):
    items = []
    for (dirpath, dirnames, filenames) in os.walk(name):
        items = filenames
        break
    matches = [m.group(1) for l in items for m in [re.search(r'^(.*\.json)$', l)] if m]
    metadata = json.loads(open(os.path.join(name, matches[0])).read())
    print(metadata)
    module = import_module(metadata['name'])
    cls = getattr(module, metadata['name'])
    cls()

if __name__ == "__main__":
    load_plugin_zip("MyPlugin.zip")
    load_plugin_directory("MyPlugin")
```

ids1024:
The regex search seems unnecessary. It would be better just to have a fixed name, like `plugin.json`.

ids1024:
I think we would be better off just including the plugins as folders. But then, what would be the best way to install them? Setup.py data files? It looks like it is not possible to include directories, based on the setuptools documentation. This could be an issue with the py2exe builds. The current version in the branch doesn't work with py2exe.

Downloading zips would be an option, but we would probably have to do it at run time rather than install time, since the same issues with data files could occur. Perhaps we can get py2exe to embed the files, but it looks like `zipimport` requires an actual path to the zip file.

lol768:
> The regex search seems unnecessary. It would be better just to have a fixed name, like plugin.json.

Fair enough, might speed things up a bit.

> Downloading zips would be an option, but we would probably have to do it at run time rather than install time, since the same issues with data files could occur. Perhaps we can get py2exe to embed the files, but it looks like zipimport requires an actual path to the zip file.

We could probably extract them from the exe to a filesystem location if the importer needs actual files. Could be done at runtime (some sort of first run flag could be cleared afterwards).

ids1024:
> We could probably extract them from the exe to a filesystem location if the importer needs actual files. Could be done at runtime (some sort of first run flag could be cleared afterwards).

That could work. We could do the same thing with the normal (non-py2exe) installations, so that they use `pkg_resources` to copy the plugins to the users plugin directory. This would improve consistency (and not require ugly py2exe specific hacks) and make it work for other installs that are not ordinary filesystem installs (I think this would be required to work with the eggs installed by easyinstall as well).

Separate repos for each plugin may not be necessary. We could probably just attach the zips to the downloads of the release on github. I am not sure which would be better.

Of course, there is always an option to have some complicated plugin manager that can install and update plugins, those included with the release as well as third party ones, built into mpsyt and downloading from a central repository. I don't think there would be much use in such a system though, at least not yet.

np1:
@ids1024 I have transferred pafy and mps-youtube into a new organisation (mps-youtube) and added you as owner.

ids1024:
I have changed the method for accessing the default plugins. The syntax is now `pluginload album_search`, although it still crashes when the command is used.

I still haven't gotten it to work for `py2exe`. I have tried using `other_resources` as described in https://stackoverflow.com/questions/1939883/py2exe-embed-static-files-in-exe-file-itself-and-access-them, but have yet to get that to work.

lol768:
Did we rule out downloading plugins on first-run? Might be a better cross-platform approach?

ids1024:
Downloading on first run is still an option, but I would prefer to avoid it. We can do that if a better method does not work. The current method should allow for running the plugins straight from the git repo during development (the importing code is a little buggy at the moment and needs some improvement, but it does work that way). Downloading at first run would require some other setup for developing the core plugins.

Base64 encoding is an option, and could work. I don't know how that would perform and what size it would take with all the extensions, however. If we don't find any issues with it, it seems like a very clean and universal solution.

I think this is only a problem with the py2exe version. Perhaps it would just be better to not have everything in a single binary, and instead have a zip that needs to be extracted containing the binary, the core plugin zip, and anything else that might be needed. I believe that is how the windows builds of mpv and mplayer need to be installed (or maybe one of them has an installer). We could optionally build a windows installer for it as well.

ids1024:
Ok, encoded base85 (smaller size than base64) in a python file fixes it under py2exe. We can keep this method is no problems arise (and they probably won't).

ids1024:
So, here is the plan for the api for holding content to be displayed now. It is important that this is well designed in advance, because a lot of mpsyt's code will then be ported to use this api.

There will be a `Content` class, which may be abstract or could have some simple default implementation. Child classes should have a `get_content()` method, which should be a generator (or other function returning an iterator) returning each item to be displayed. Results of this will be cached, and the next item retrieved from the iterator as needed.

Originally the idea was for these items to be text lines ready to be displayed, but that won't work so well for code that adjusts to window width, etc. So they could be in any format, but another method is needed to process them, the `format_item` method (if this sounds terrible, please do suggest a better design).

That function takes the item retrieved from the iterator and the width of the screen (other parameters as well?), and should return a string to be displayed on the screen.

There will also be a `TabularContent` class, a subclass of `Content`, for displaying content in tables, as is often used in mpsyt.

I am still not confident with this design. For one thing, tables could be displayed better using other methods in e.g. an urwid-based curses interface, which would not be possible using this purely text-based method.

Also, there are another two optional methods of `Content` subclasses, the `get_length()` method, to return the number of items if that is known in advance, and the `get_message()` method, for a message to display at the bottom of the screen if needed.

Any comments or suggestions?

ids1024:
I need to get back to this. I think perhaps the interface for defining commands could be improved. Instead of a single regex, it could have the command and parameters specified as function parameters, which were added to python in version 3 (which mpsyt requires, so that is not an issue). Something like this:

    @commands.command
    def thecommand(command: r"(thecommand)\s+", param1: r"(\S+)\s*$"):

Perhaps spaces should be handled by the software.

    @commands.command
    def thecommand(command: r"thecommand", param1: r"\S+")

And the default, if none is given, can be `r"\S+"` (any string more than 1 char without whitespace)

    @commands.command
    def thecommand(command: r"thecommand", param1):

But then, assuming the current command interface is kept, there must be a way to not require spaces.

    @command.command(add_spaces=False)
    def search(command: r"(/)", term):

Or is there a better way? I like this method, except perhaps the handling of commands which do not have a whitespace after the command or which do not use spaces between parameters (like `1-6` or other variants of playing tracks.) I suppose the second case could just be handled as a single parameter then split by the function (it will already have confirmed it fits the regex). Anyone else have any ideas on how this could be improved?

ids1024:
The-Compiler/qutebrowser#30 contains a good amount of research into how best to design a plugin API for python software. I might steal some ideas from there.

Razesdark:
I'm just going to close down this PR for now.

PR #336, opened 2015-08-02
Changed the mplayer version regex

Evidlo:
Changed the mplayer version regex so that the following two Mplayer version strings match successfully

`MPlayer svn r34540 (Debian), built with gcc-4.7 (C) 2000-2012 MPlayer Team`
`MPlayer SVN-r37391-5.1.1 (C) 2000-2015 MPlayer Team`

Version strings are from distro-packaged mplayer on Debian 8 and Fedora 22, respectively.


Evidlo:
And yes, I'm aware that mplayer in the first version string is too old.

PR #338, opened 2015-08-05
catch all dbus exception in mpris

### LINKED ISSUE #337: Dbus error
Today mpsyt begins with this error:

>File "/usr/lib/python3.4/multiprocessing/process.py", line 254, in _bootstrap
    self.run()
  File "/usr/lib/python3.4/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
Enter /search-term to search or [h]elp
  File "/usr/local/lib/python3.4/dist-packages/mps_youtube/mpris.py", line 519, in main
    mprisctl.acquire()
  File "/usr/local/lib/python3.4/dist-packages/mps_youtube/mpris.py", line 72, in acquire
    self._acquire_bus()
>   File "/usr/local/lib/python3.4/dist-packages/mps_youtube/mpris.py", line 113, in _acquire_bus
    bus=dbus.SessionBus(mainloop=DBusGMainLoop()))
  File "/usr/lib/python3/dist-packages/dbus/_dbus.py", line 211, in __new__
    mainloop=mainloop)
  File "/usr/lib/python3/dist-packages/dbus/_dbus.py", line 100, in __new__
    bus = BusConnection.__new__(subclass, bus_type, mainloop=mainloop)
  File "/usr/lib/python3/dist-packages/dbus/bus.py", line 122, in __new__
    bus = cls._new_for_bus(address_or_type, mainloop=mainloop)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.NotSupported: Unable to autolaunch a dbus-daemon without a $DISPLAY for X11

When play a sing starts, but after that  the player, mpv or mplayer stop, and I have to cancel the player. The next time I play a song never start. 

This hapen only in tty, xterm works fine.

Today use python enviroment (venv) to install a pip package, using 2.7, I don't now if the venv affect mpsyt in tty.

hrnr:
since mpris is not critical, rest of the mps-youtube can run with no problems

fixes #337

### DIFF
--- a/mps_youtube/mpris.py
+++ b/mps_youtube/mpris.py
@@ -516,6 +516,10 @@ def main(connection):
     """
 
     mprisctl = Mpris2Controller()
-    mprisctl.acquire()
-    mprisctl.run(connection)
-    mprisctl.release()
+    try:
+        mprisctl.acquire()
+        mprisctl.run(connection)
+    except dbus.exceptions.DBusException:
+        pass
+    finally:
+        mprisctl.release()
--- a/mps_youtube/mpris.py
+++ b/mps_youtube/mpris.py
@@ -518,8 +518,8 @@ def main(connection):
     mprisctl = Mpris2Controller()
     try:
         mprisctl.acquire()
-        mprisctl.run(connection)
     except dbus.exceptions.DBusException:
-        pass
-    finally:
-        mprisctl.release()
+        print('mpris interface couldn\'t be initialized. Is dbus properly configured?')
+        return
+    mprisctl.run(connection)
+    mprisctl.release()


ids1024:
It should at least catch when there is no dbus server running, but I don't know about catching every error. Perhaps for other dbus issues it should print a warning and write to the log (if logging is enabled) or at the very least write to the log.

hrnr:
agree. logging is always good. What do we have for logging? I now realised I maybe made this in a hurry, I will look on that again. We probably want to continue on some errors (like bad response) and return on others.

ids1024:
Exeception handling in mpsyt could currently use improvement in other parts as well. Some debugging involves deleting except statements to get an actually useful error message, which is not good. This could be the same if there is some kind of issue with the mpris code, not that there have been any like that yet.

hrnr:
Sorry for delay. I took a look on error reporting in mps-youtube and mostly we just prints to terminal. I have taken the same approach.

I think it is sufficient to catch errors only on initialization. As we don't have a proper logging facility minimal approach is probably better.

PR #339, opened 2015-08-05
do not wait for notifier

hrnr:
this allows notifier to do some more complex operations without causing delay in playback.


ids1024:
@hrnr Just out of curiosity, are you doing something interesting with a notifier that required this change, or is it just something random you noticed that could be improved?

PR #349, opened 2015-09-08
Use terminal control sequences only on terminals

livingBEEF:
This makes it possible to use pagers with --help.


ids1024:
Ok, looks reasonable enough.

PR #372, opened 2015-11-03
allowed to specify youtube-dl options from mps-yt

kkrolczyk:
MPS had been changed minimally, by adding 2 checking functions to Config, both modify global dictionary ydl_opts in 'g' module. Also pafy.new() call was updated. 
This patch requires that pafy accepts 'ydl_opts' as argument to 'new' function, in case mps-yt was updated and pafy was not - current design in pafy would cause it to break (TypeError, pafy.new() got unexpected keyword argument ydl_opts)

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -361,7 +361,6 @@ class _Config(object):
             ConfigItem("download_command", ''),
             ConfigItem("api_key", "AIzaSyCIM4EzNqi1in22f4Z3Ru3iYvLaY8tc3bo",
                 check_fn=check_api_key),
-            ConfigItem("yt_dl_use_https", False, check_fn=update_use_https),
             ConfigItem("yt_dl_cmdline", fake_string("<empty>",{}),
                 check_fn=check_ydl_opts_valid),
             ]
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -226,6 +226,38 @@ def check_win_size(size):
     else:
         return dict(valid=True, value=size)
 
+def check_being_dict(user_dict):
+    """ check if user string is correct dict, set it to global options,
+        since pafy accepts ydl_opts dict as ctor argument. """
+    try:
+        import ast
+    except ImportError:
+        m = "Your Python is too old. Use at least 2.4 to use this option."
+        return dict(valid=False, message=m)
+    try:
+        dict_ = ast.literal_eval(str(user_dict))
+    except ValueError:
+        m = "Entered string has wrong syntax. Usage: {'option': value}"
+        return dict(valid=False, message=m)
+    if isinstance(dict_, dict):
+        # it is assumed that user setting this option knows youtube-dl options
+        # names (like:quiet, prefer_insecure, no_warnings, etc)
+        g.ydl_opts = dict_
+        return dict(valid=True, message="Youtube-dl options set successfully")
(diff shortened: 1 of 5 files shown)


mps_youtube/config.py
```diff
@@ -225,6 +225,38 @@ def check_win_size(size):
     else:
         return dict(valid=True, value=size)
 
+def check_being_dict(user_dict):
+    """ check if user string is correct dict, set it to global options,
+        since pafy accepts ydl_opts dict as ctor argument. """
+    try:
+        import ast
+    except ImportError:
+        m = "Your Python is too old. Use at least 2.4 to use this option."
```
ids1024: Mysyt no longer supports (nor functions in) any version of python 2, so this test should be unnecessary.

ids1024:
Maybe it would be better to use a different format for specifying arguments. Using the same format as normal command line arguments would probably be more user friendly and intuitive, if that would not be too complicated to implement.

I'm not quite sure about the best way to do that though. Maybe by directly calling `youtube_dl.options.parseOpts`, but that would probably result in undesired side effects (it looks like it prints some strings if the debug option is set, and probably displays a help message if `--help` is used). And it might not be safe to rely on that as a stable api.

kkrolczyk:
@ids1024 hmm. It might be better, yet I think it would mean either massive code duplication (since yt-dowloader already has parsing and detecting what kind of options they support, i think it would be doable to use their argparsers - like You mentioned) or own parsing, which in turn would inflate code quite a bit (single, double dash, and god knows [how many args they support] (https://github.com/rg3/youtube-dl/blob/master/youtube_dl/YoutubeDL.py#L121), not to mention validation) - it is doable, and not too complicated I suppose; however it would bloat, require synchronization with youtube-dl, and frankly I'm not sure, would be really more intuitive? After all this setting would need to be converted to dictionary eventually. But - maybe You're right. I'll think about it, and I'll give a try if this is ugly hack or not. And regarding user passing `--help` to mps-yt internal settings...do You think people would really want to do that? :) I mean sure, args can filter-out but...i'm just hoping people using this aren't mindless, and suppose it might be even a little impolite to suggest so ;)

As a side note - I have been wondering how much of backwards compatibility do You want to preserve? I've been  thinking, for example, instead of passing options to new(), why not pass *args, **kwargs directly to Pafy class? 

**tl;dr:**
I think using their arg parser is good idea; as to api stability - since pafy depends on yt-dl I'm not sure how this could be a disadvantage? Regarding noise from youtube_dl.options.parseArgs, maybe stdout could be captured and redirected for a moment to null/stringIO?
Anyway I'll check if parsing such setting formated as cmd args is sane. I haven't been using yt-dl from cmd, could You propose some switches for tests?

ids1024:
Yes, use their arg parser as long as it does not cause any problems. Backwards compatibility is not an absolute necessity, but in this case, it is fairly important. Currently (due to some argument about the merits of using youtube-dl as a backend; see mps-youtube/pafy#110), a 0.3.x branch of pafy is, at least for now, being maintained with bug fixes but without the change to youtube-dl. So that branch needs to be supported too.

kkrolczyk:
Ok, i think this is as good as it gets, unless someone can come up with better idea. This allows mps-yt to set and store youtube-dl commandline arguments in its settings (using previously added pafy functionality). I did some preliminary testing, but I'd be happy if second pair of eyes could check if i missed some corner cases.

Issue #360, opened 2015-10-12
[Request] Don't save numbers and hotkeys in the mpsyt search history.


GijsTimmers:
Current situation:

```
$ cat ~/.config/mps-youtube/input_history
/mauro picotto proximus
r1
/barthezz on the move
1
q
/then came you
1
/pursuit of happiness
1
2
q
```

Entries like `1`, `2`, `r1`, `q` are used to indicate what track to play. They end up in the search history (<kbd>Ctrl</kbd> + <kbd>r</kbd>), which makes it harder to find out what tracks I looked up in the past.

Request: don't save these numbers and hotkeys, or create a special command like `history` (currently still available) to show the relevant search history instead.

kkrolczyk:
It could be implemented for example like this:
https://github.com/mps-youtube/mps-youtube/commit/61a6ef173ade1445ecd09041516c050c064fa391
yet, it appears that mpsyt's Config.\<item\>.get in main.py gets default value instead of such that was previously set in config.

lol768:
:+1: to this

ids1024:
> yet, it appears that mpsyt's Config.<item>.get in main.py gets default value instead of such that was previously set in config. 

What? I don't think that can be the case. If it were, none of the settings would do anything. What gives you that impression?

kkrolczyk:
>> yet, it appears that mpsyt's Config..get in main.py gets default value instead of such that was previously set in config.

> What? I don't think that can be the case. If it were, none of the settings would do anything. What gives you that impression?

@ids1024, here is how i had it checked:

0. # this steps are shown in config_test.sh with my output
1. cloned mps-yt to brand new directory,  cd that directory
2. removed old config dir ~/.config/mps-youtube
3. git 'apply'-ied patch history_patch.patch (could have also pulled my fork and switch to branch history_fix, but i wanted to make sure that all's done properly)
4. launched new_directory/mpsyt 
5. used set to check current setting, and next set to change it to searches_only.
6. verified that setting persists ie, it is displayed as being set to proper value in next mpsyt runs
7. verified that code path `elif Config.SAVE_HISTORY.get == "searches_only":` is never taken, always `if Config.SAVE_HISTORY.get == "all":` is executed. Changed config.py default value to searches_only and code path `elif Config.SAVE_HISTORY.get == "searches_only":` was taken. also, take a look at even simplier patch to show this:

8. prepared even simpler test:

 - first, checkout previous patch, ie go back to clean develop branch:
[...]

PR #377, opened 2015-11-11
History fix issue 360

kkrolczyk:
This patch checks, during mpsyt startup, in init_readline() function if user wants different history behavior. Config.`save_history` can be set to "all" (current and previous version default behavior), none, or <nothing>, which disables readline, and "searches_only" to filter readline file during each mpsyt startup, leaving only items that were searched for, as described in feature request #360

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -68,9 +68,6 @@ class ConfigItem(object):
 
         # handle known player not set
 
-        # sadly, allowed_values is faulty - assumes copy() is always
-        # accessible, causing exception on any string not among allowed_values
-
         if self.allowed_values and value not in self.allowed_values:
             fail_msg = "%s must be one of * - not %s"
             allowed_values = copy.copy(self.allowed_values)
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -68,6 +68,9 @@ class ConfigItem(object):
 
         # handle known player not set
 
+        # sadly, allowed_values is faulty - assumes copy() is always
+        # accessible, causing exception on any string not among allowed_values
+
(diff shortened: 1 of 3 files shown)


mps_youtube/config.py
```diff
@@ -67,6 +67,9 @@ def set(self, value):
 
         # handle known player not set
 
+        # sadly, allowed_values is faulty - assumes copy() is always
+        # accessible, causing exception on any string not among allowed_values
```
ids1024: Could you explain what you mean by this?

kkrolczyk: Sure, please check  what happens if user didn't read `help config` careful, or made a typo like this
`set save_history search_onl` (notice missing 'y'). Exception is generated like this:

```
> set save_history search_onl
  File "./mpsyt", line 3, in <module>
    mps_youtube.main.main()
  File "/tmp/mps-youtube/mps_youtube/main.py", line 3819, in main
    if matchfunction(k, v, userinput):
  File "/tmp/mps-youtube/mps_youtube/main.py", line 3716, in matchfunction
    func(*matches)
  File "/tmp/mps-youtube/mps_youtube/main.py", line 471, in setconfig
    message = Config[key.upper()].set(val)
  File "/tmp/mps-youtube/mps_youtube/config.py", line 75, in set
    allowed_values = self.allowed_values.copy()
AttributeError: 'list' object has no attribute 'copy'
```

also happens when user tries to change user_order and other parameters that accept multiple values.

ids1024: Ah, now I see. I thought `.copy()` was added in python 3.0, but apparently it was added in python 3.3. d5874380d2be34103be7c019c472a3e16f456020 should fix it.

PR #385, opened 2015-11-22
Fix right-aligned status line prompt when using colors.

JKatzwinkel:
Apply regex to calculate actual status line length regardless of color codes.

### DIFF
--- a/mps_youtube/c.py
+++ b/mps_youtube/c.py
@@ -34,13 +34,13 @@ else:
 
 r, g, y, b, p, w = red, green, yellow, blue, pink, white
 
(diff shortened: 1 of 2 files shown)


ids1024:
Oh, yeah, this does need to be change.

But why `mlen()`? I don't see where the `m` comes from (although maybe it is more obvious than it seems to me). Perhaps a more descriptive function name would be better.

JKatzwinkel:
I'm really bad at naming things and I don't recall what the `m` stands for either. I suspect 'monochrome', but it has been a while since then and I can't be sure. Anyway I try to come up with a better name.

Issue #355, opened 2015-09-28
[Feature Request] Open in default web browser


maricn:
Hi guys,

Do you think it would be useful to have a command (or button during playback) which would open YouTube URL in default (or configurable) web browser?
I often want to save a video to my account playlist or to subscribe to a channel, or whatever, but I must repeat a few steps until I get there. I figure this isn't so hard to implement...

Wildefyr:
This would be a nice stop gap solution to the fact we cannot log into our google youtube accounts with mpsyt.

BensonQiu:
Hi,

[I'm](https://github.com/BensonQiu) working on a group project with [Femi](https://github.com/aquafemi), [Paul](https://github.com/paulnicolet), [Rohan](https://github.com/rjvani), and [Tommy](https://github.com/tommyhtc), and we're interested in contributing to mps-youtube. We don't have much experience with open source contributions, but we would like to work on [#90](https://github.com/mps-youtube/mps-youtube/issues/90), [#375](https://github.com/mps-youtube/mps-youtube/issues/375), and [#355](https://github.com/mps-youtube/mps-youtube/issues/355) as a group over the next 2-3 weeks. Would these contributions be helpful for the project?

We have read the [contributing](https://github.com/mps-youtube/mps-youtube/blob/develop/CONTRIBUTING) document, but we wanted to post here to express interest. We also wanted to see if anyone was already working on these issues. Since we would like to work on three open issues, I'm posting this message three times (sorry for the spam!).

Regards,
Benson, Femi, Paul, Rohan, and Tommy

Wildefyr:
Just thought I'd mention in addition to anyone wanting this at the moment, it's possible to get url recognition inside the terminal itself, i.e. urxvt-perls provides extensions for this functionality for rxvt-unicode. You can then see the youtube url with the `url #` command in mpsyt.

PR #390, opened 2015-12-03
Open YouTube URL in default browser

rjvani:
Added functionality to open YouTube URLs and IDs in the default browser. #355

### DIFF
--- a/mps_youtube/helptext.py
+++ b/mps_youtube/helptext.py
@@ -79,7 +79,7 @@ def helptext():
     {2}dvupl <username>{1} - download user's YouTube playlists (video).
     {2}dlurl <url or id>{1} download a YouTube video by url or video id.
     {2}playurl <url or id>{1} play a YouTube video by url or id.
-    {2}browserplay <url or id>{1} play a YouTube video by url or id in browser.
+    {2}browserplay <number>{1} open a specified previous search in browser.
     {2}browsersearch <term(s)>{1} open search results for term(s) in browser.
 
     {2}all{1} or {2}*{1} - play all displayed items.
--- a/mps_youtube/helptext.py
+++ b/mps_youtube/helptext.py
@@ -80,7 +80,6 @@ def helptext():
     {2}dlurl <url or id>{1} download a YouTube video by url or video id.
     {2}playurl <url or id>{1} play a YouTube video by url or id.
     {2}browserplay <number>{1} open a specified previous search in browser.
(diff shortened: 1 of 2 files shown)


bendem:
That doesn't allow opening a search result in the browser right?

rjvani:
Correct. This opens a link to the video directly given an ID or URL. I can work on that part as well, though.

rjvani:
I've pushed changes to allow for opening search results in the browser.

mps_youtube/helptext.py
```diff
@@ -79,6 +79,8 @@ def helptext():
     {2}dvupl <username>{1} - download user's YouTube playlists (video).
     {2}dlurl <url or id>{1} download a YouTube video by url or video id.
     {2}playurl <url or id>{1} play a YouTube video by url or id.
+    {2}browserplay <url or id>{1} play a YouTube video by url or id in browser.
```
ids1024: Wouldn't it make sense for this to accept numbers as well? Like `browserplay 1` to open the first result in the browser.

rjvani: browserplay was intended for opening a given ID or URL. Since it points to a single video, I'm not really sure how the number would help. I can work on adding this functionality to browsersearch, though, which opens the search query in the browser.

bendem: I thought the requested feature was to search a video from mpsyt and then open the video inside your browser like
```
/the nights
browser 1-2
```
which would open the results 1 and 2 of 'the nights' search in the browser.

ids1024: I mean that it should do something like @bendem said, if that isn't clear.

That seems to be where this is particularly useful. For an url, this is fairly useless, since you can just enter it into the browser directly. Handling an id is mildly more useful, but usually you get one from looking at the url anyway. Taking a number for a search result, playlist item, etc. is where it is useful, and integrates with mpsyt's functionality.

rjvani: Oh okay, my bad! I'll update my function to allow for that. :)

rjvani: I've pushed changes so that browserplay can open a certain video from previous search instead of a URL or id.

ids1024:
I don't know, is the `browsersearch` command really useful? I don't that's what @bendem meant (correct me if I'm wrong), but rather the functionality you just added.

rjvani:
I might have misunderstood that initially. I'll remove that and leave the most recent functionality. My bad!

bendem:
Awesome, thanks for the patch.

Issue #90, opened 2014-06-13
Create playlist from list of urls


bburky:
Is it currently possible to create a playlist from a list of urls?

Manually it would be possible with:

```
url https://www.youtube.com/watch?v=...
add 1
url https://www.youtube.com/watch?v=...
add 1
...
```

But this is tedious. My suggestion is to allow the `url` command to accept multiple urls.

I would like to pass in a list of urls from the command line, which this would make possible. I could run
`mpsyt url https://www.youtube.com/watch?v=... https://www.youtube.com/watch?v=... https://www.youtube.com/watch?v=...` or
`mpsyt url $(generate list of urls somehow)`.

Alternately you could read a list of urls from a file. Or allow multiple commands to be run from the command line.

BensonQiu:
Hi,

[I'm](https://github.com/BensonQiu) working on a group project with [Femi](https://github.com/aquafemi), [Paul](https://github.com/paulnicolet), [Rohan](https://github.com/rjvani), and [Tommy](https://github.com/tommyhtc), and we're interested in contributing to mps-youtube. We don't have much experience with open source contributions, but we would like to work on [#90](https://github.com/mps-youtube/mps-youtube/issues/90), [#375](https://github.com/mps-youtube/mps-youtube/issues/375), and [#355](https://github.com/mps-youtube/mps-youtube/issues/355) as a group over the next 2-3 weeks. Would these contributions be helpful for the project?

We have read the [contributing](https://github.com/mps-youtube/mps-youtube/blob/develop/CONTRIBUTING) document, but we wanted to post here to express interest. We also wanted to see if anyone was already working on these issues. Since we would like to work on three open issues, I'm posting this message three times (sorry for the spam!).

Regards,
Benson, Femi, Paul, Rohan, and Tommy

ids1024:
@BensonQiu Contributions are of course welcome. It probably would have been best to only post the message once though, since github shows a link when an issue is referenced in another issue, making the duplication a little redundant. But anyway...

This issue is not being worked on, though some thought has gone into #361 (as you can see there) which is somewhat related. Nor is there presently any work done on #374 and #355. So if you would like to implement these, go ahead.

PR #391, opened 2015-12-04
Import videos from a list of urls or ids, from the console or .txt file.

paulnicolet:
This commit aims to implement the features requested in #90.
You can know type several URL's or ID's with the `url` command, or use a new command `url_file` to import this list from a .txt file.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -3293,21 +3293,14 @@ def yt_url_file(file_name):
     #Open and read the file
     try:
         fo = open(file_name, "r")
-        lines = fo.readlines()
     except (IOError):
-        g.message = c.r + 'Error while reading the file' + c.w
+        g.message = c.r + 'Error while opening the file, check the validity of the path' + c.w
         g.content = g.content or generate_songlist_display(zeromsg=g.message)
         return
 
-    output = ""
-    for line in lines:
-        #Don't forget to remove the \n character
-        if line[-2:] == '\\n':
-            line = line[:-2]
-        output += (line + ',')
+    output = ','.join([line.strip('\n') for line in fo])
 
-    #Remove the final coma
-    output = output[:-1]
+    fo.close()
 
     #Finally pass the input to yt_url
     yt_url(output)
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -3292,15 +3292,13 @@ def yt_url_file(file_name):
 
     #Open and read the file
     try:
-        fo = open(file_name, "r")
+        with open(file_name, "r") as fo:
+            output = ','.join([line.strip('\n') for line in fo])
+
     except (IOError):
         g.message = c.r + 'Error while opening the file, check the validity of the path' + c.w
         g.content = g.content or generate_songlist_display(zeromsg=g.message)
         return
-
-    output = ','.join([line.strip('\n') for line in fo])
-
-    fo.close()
 
     #Finally pass the input to yt_url
     yt_url(output)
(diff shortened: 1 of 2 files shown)


mps_youtube/main.py
```diff
@@ -3280,6 +3287,31 @@ def yt_url(url, print_title=0):
     if print_title:
         xprint(v.title)
 
+def yt_url_file(file_name):
+    """ Access a list of urls in a text file """
+
+    #Open and read the file
+    try:
+        fo = open(file_name, "r")
```
BensonQiu: It looks like `fo` is never explicitly closed. 
Perhaps try `with open(file_name, "r") as fo:`?

mps_youtube/main.py
```diff
@@ -3280,6 +3287,31 @@ def yt_url(url, print_title=0):
...
+        lines = fo.readlines()
+    except (IOError):
+        g.message = c.r + 'Error while reading the file' + c.w
+        g.content = g.content or generate_songlist_display(zeromsg=g.message)
+        return
+
+    output = ""
```
BensonQiu: Just a minor style suggestion - you can replace lines 3302 - 3310 with:
```
output = ','.join(
    [line.strip('\n') for line in lines]
)
```

mps_youtube/main.py
```diff
@@ -3280,6 +3287,31 @@ def yt_url(url, print_title=0):
...
+    for line in lines:
+        #Don't forget to remove the \n character
+        if line[-2:] == '\\n':
```
ids1024: I don't think this is correct. `\\` would make it a literal backslash character.

It would probably be better to use `line = line.strip()`.

mps_youtube/main.py
ids1024: You can also use `[line.strip('\n') for line in fo]`, instead of calling `fo.readlines()`.

paulnicolet:
Thanks for the style tips. I committed the changes.

paulnicolet:
Thanks for the style tips. I applied the changes.

mps_youtube/main.py
```diff
@@ -3746,6 +3769,7 @@ def main():
         quits: r'(?:q|quit|exit)$',
         plist: r'pl\s+%s' % pl,
         yt_url: r'url\s(.*[-_a-zA-Z0-9]{11}.*$)',
+        yt_url_file: r'url_file\s([^.]+\.txt$)',
```
bendem: I'd expect this feature to work with any file extension, not just `.txt`. Maybe something like 
```py
r'url_file\s+([^\s]+$)'
```

mps_youtube/main.py
```diff
@@ -3280,6 +3287,22 @@ def yt_url(url, print_title=0):
     if print_title:
         xprint(v.title)
 
+def yt_url_file(file_name):
+    """ Access a list of urls in a text file """
+
+    #Open and read the file
+    try:
+        with open(file_name, "r") as fo:
+            output = ','.join([line.strip('\n') for line in fo])
```
bendem: File format would be more flexible with 
```py
[line.strip() for line in fo if line.strip()]
```
(allows for empty lines and strip all types of whitespaces)

ids1024:
I merged this, but changed the syntax a bit (see cd308039883edf38b440d5b53cb28d7915d61578). I made the urls delimited by whitespace instead of comma in the `url` command, and made it so the file given to `url_file` can have any extension.

mps_youtube/main.py
ids1024: @bendem Oh, I didn't see that comment before merging, but I agree: cccbf0894172a738732715c1f8f876b0992e49f5

PR #394, opened 2015-12-14
remove unneeded showconfig() argument

Wildefyr:
Not sure what the intention behind this was but keeping it as it is crashes mps-youtube.


ids1024:
Oh, thanks. Showconfig used to require an argument (which was ignored), but that was changed recently.

Issue #201, opened 2015-01-25
Limited results for `all`


tarwich:
When playing all songs from a YouTube user, it would be nice if the system could somehow play all the songs instead of limiting you to the hight of the terminal window.

**Steps to reproduce**
 - Run `mpsyt`
 - Type `user smoothmcgroove`
 - Type `all`

**Expected result**
Should play all 70+ videos for user

**Actual result**
Limits playback entries to height of terminal window

np1:
Yep that would be useful, the `dump` command could be made to work with user results as well as YouTube playlists.

bendem:
Would it be much more work than adding `g.last_search_query.get("user")` [here](https://github.com/np1/mps-youtube/blob/develop/mps_youtube/main.py#L3971)?

PR #399, opened 2015-12-23
[WIP] Use paginatesongs() for all search commands

ids1024:
This is a step forward to having a better internal api that doesn't require directly setting globals, which is a prerequisite for the plugins api (#314 #316).

~~Note: album search is currently broken on this branch. It is... problematic.~~ Fixed


ids1024:
This implements #201 as well.

Issue #397, opened 2015-12-23
I have to issue 'set player mpv' every time on Windows


GijsTimmers:
Using Windows 7 x86.

Some debug information:

`mpsyt`:
```
C:\Users\user>C:\Python34\Scripts\mpsyt.exe --version
mpsyt version  : 0.2.5
   notes       : released 1 June 2015
pafy version   : 0.4.2
Python version : 3.4.3 (v3.4.3:9b73f1c3e601, Feb 24 2015, 22:43:06) [MSC v.1600
32 bit (Intel)]
Processor      : x86 Family 6 Model 23 Stepping 10, GenuineIntel
Machine type   : x86
Architecture   : 32bit, WindowsPE
Platform       : Windows-7-6.1.7601-SP1
sys.stdout.enc : cp850
default enc    : utf-8
Config dir     : C:\Users\user\AppData\Roaming\mps-youtube
```

`mpv.com`:
```
C:\Users\user>C:\Python34\Scripts\mpv.com --version
mpv 0.14.0-git-e1993d5 (C) 2000-2015 mpv/MPlayer/mplayer2 projects
 built on Fri Dec 11 20:32:28 CET 2015
ffmpeg library versions:
   libavutil       55.10.100
   libavcodec      57.17.100
   libavformat     57.19.100
   libswscale      4.0.100
   libavfilter     6.20.100
   libswresample   2.0.101
ffmpeg version: N-77205-gcc4718a
```

`set` output:
```
  Key                 Value
  api_key           : sOmeApIkEy
  checkupdate       : True
  colours           : True
  columns           :
  ddir              : C:\Users\user\Downloads\mps
  download_command  :
  encoder           : 0 [None]
  fullscreen        : False
  max_res           : 2160p
  notifier          :
  order             : relevance
  overwrite         : True
  player            : mpv.exe
  playerargs        :
  search_music      : True
  show_mplayer_keys : True
  show_status       : True
  show_video        : True
  user_order        :
  window_pos        :
  window_size       :

Enter set <key> <value> to change
Enter set all default to reset all
```

The problem I'm facing:

I start `mpsyt`. I look up a video. `mpsyt` gives the expected search results. When trying to play a video, I'm getting `error: retrying` in the bottom left corner of the terminal.

The strange thing: after issuing `set player mpv` and choosing the desired video again, the video plays.

I exit `mpsyt`, and start it again. The same problem arises: I have to issue `set player mpv` again before I can play videos.

[...]

PR #411, opened 2016-01-10
proof-of-concept fix for #397

GijsTimmers:


ids1024:
Yes, that should work, but it would be simpler just to put "mpv" directly in the regex string.

I did that in 4ba1d41e12624c4df1d5b57a0fba2e9cf75bab78, so it should work now.

Issue #415, opened 2016-01-15
Issue when setting playerargs with mpv


sotlampr:
I am trying to set playerargs for mpv, let's say I want the video output to libcaca so I must set 

    --vo=caca

I see two ways of doing this:
1. With quotes:

       set playerargs "--vo=caca"

(or the equivalents "-vo=caca", -"-vo caca", "-vo caca" etc.)

What happens is that mpv thinks the arguement is a file, so I have the message "cannot open file "--vo=caca" if I omit the --really-quiet option and, of course, no libcaca output.

2. Without quotes:

        set playerargs --vo=caca

(or the equivalents -vo=caca, --vo caca, -vo caca etc.)

What is happening now is that the video plays as supposed,  i.e. with libcaca output (if i freeze at this point with pdb), BUT after opening the mpv subprocess an error is raised when sending the socket to mprisctl ([main.py, L1262](https://github.com/mps-youtube/mps-youtube/blob/develop/mps_youtube/main.py#L1261)). The error states:

*Problem playing last item: Expecting property name enclosed in double quotes: line 1 column 62 (char 61)*

Any Ideas or suggestions on how to fix this?

S.

ids1024:
I can't seem to reproduce this. Try setting a breakpoints at `mps_youtube.main.launch_player` and using `args` to see exactly what command it is trying to execute.

sotlampr:
OK, to recap: when using playerargs --vo=caca (without quotes) I get the following errror:

    Problem playing last item: Expecting property name enclosed in double quotes: line 1 column 62 (char 61)

The error comes from [mpris.py, L212](https://github.com/mps-youtube/mps-youtube/blob/develop/mps_youtube/mpris.py#L212) and is this:

    File "./mps-youtube/mps_youtube/mpris.py", line 214, in bindmpv
      resp = json.loads(line)
    File "/usr/lib/python3.5/json/__init__.py", line 319, in loads
      return _default_decoder.decode(s)
    File "/usr/lib/python3.5/json/decoder.py", line 339, in decode
      obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    File "/usr/lib/python3.5/json/decoder.py", line 355, in raw_decode
      obj, end = self.scan_once(s, idx)
    json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: 
    line 1 column 62 (char 61)

the offending line is:

    {"event":"property-change","id":1,"name":"time-pos","data":0,000000}
[...]

PR #417, opened 2016-01-20
Fix for issue #415 for non-english locales

sotlampr:
As discussed in [#415](https://github.com/mps-youtube/mps-youtube/issues/415) and after [this response](https://github.com/mpv-player/mpv/issues/2740) from the mpv team, I think this fix is the sanest solution. Fixes the issue, passes the tests on linux.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -95,7 +95,7 @@ except ImportError:
 mswin = os.name == "nt"
 not_utf8_environment = mswin or "UTF-8" not in sys.stdout.encoding
 
-locale.setlocale(locale.LC_ALL, "")  # for date formatting
+locale.setlocale(locale.LC_TIME, "")  # for date formatting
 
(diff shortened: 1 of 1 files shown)


sotlampr:
Sorry, accidentally added newline

ids1024:
This should be fine, but it is a bit of a hack. If the problem is with libcaca, perhaps it would be good to report it to their bug tracker and see if the problem can be fixed there?

hrnr:
Does it even fix your problem? As I currently understand the situation, locale that we set in python have nothing to do with locale mpv uses. (We run it as a standalone process.)

As mentioned in mpv-player/mpv#2740, it seems libcaca is setting locale for mpv (unfortunately).

sotlampr:
I checked again and you are right, it doesn't fix my problem, I somehow mixed up locale configuration. I'm going to forward the issue to libcaca.
Thanks!

PR #424, opened 2016-02-01
Play an entire YouTube playlist command

danickfort:
* Allows playing an entire opened YT playlist with
    the double asterisk command "**".
    It essentially is a shortcut for the sequence:
      dump, *

### DIFF
--- a/mps_youtube/helptext.py
+++ b/mps_youtube/helptext.py
@@ -86,6 +86,7 @@ def helptext():
     {2}browserplay <number>{1} - open a specified previous search in browser.
 
     {2}all{1} or {2}*{1} - play all displayed items.
+    {2}**{1} - play all items on every page.
     {2}repeat <number(s)>{1} - play and repeat the specified items.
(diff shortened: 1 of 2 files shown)


ids1024:
I think it might be best to make this the behavior of `*`/`all`, and get rid of the `dump` command. But that requires some other changes, such as loading additional pages in the background so it doesn't take forever to start playing. It would also make sense to have numbering start from the first page instead of restarting from one on every page.

I have been intending to do this, but it would take some time to implement. It would probably make sense to do after #423 which I have recently been working on is finished, since both involve replacing `g.model` with something else.

danickfort:
Alright, sounds good. I'll close this for now then. Do you need some help on that new content API ? I'd love to help develop this project further.

ids1024:
@danickfort Sure, help would definitely be appreciated.

ids1024:
The dump command is now gone in the `content` branch, with behavior as I mentioned above. That branch is quite broken, and not really usable, but hopefully will be finished and merged before all too long (famous last words).

Issue #466, opened 2016-03-31
Select default audio format + quality in config | Feature request


nishanthkarthik:
Currently, `mpsyt` opens `webm` for some links and `m4a` for others. Can we have an option in set option to select default format and quality?

It would be wonderful when this happens :)

ids1024:
Selecting the format has been requested before: #71 
It's shouldn't be hard to implement... feel free to implement it and send a PR.

As for quality, you can set the maximum resolution for video playback with `set max_res`. For audio it just uses the highest bitrate. I suppose a `max_bitrate` setting could be added too... Though it seems much less useful that `max_res`.

nishanthkarthik:
Thanks @ids1024 
This is my first contribution to a large scale open source project :)

PR #467, opened 2016-03-31
Add settings for default audio format

nishanthkarthik:
Referencing issue  #466 

I have added `m4a` and `webm` as options for `audio_format` configuration


spvkgn:
@nishanthkarthik Can you please add the same option for video format (webm/mp4) ?

Issue #344, opened 2015-08-24
Feature request: persistent history


jrenner:
Sometimes I listen to a song and forget the name or search terms I used to find it. It would be nice if there was a persistent history that I could search through to try and find it again.

sees-no-colors:
+1

maricn:
+1 from me too

PR #472, opened 2016-04-10
Add a history of played songs

kraetzin:
Initial implementation. Referencing issue #344.

I implemented a function `history_add(song)` which is called each time `playsong()` is successfully called. It adds the song to a history playlist which is implemented separately from the main playlist file via the functions `open_hist_from_file()` and `save_to_hist()`. The history file is `play_history`. This can be viewed with the `history` command. I also implemented `history clear` to clear the history. I'm holding off on adding a more rigorous history implementation due to this being my first contribution


JKatzwinkel:
Nice!

mps_youtube/main.py
```diff
@@ -2137,6 +2182,17 @@ def playlist_add(nums, playlist):
     g.content = generate_songlist_display()
 
 
+def history_add(song):
+    """ Add song to history. """
+    if not g.userhist.get('history'):
+        g.userhist['history'] = Playlist('history')
+
+    g.userhist['history'].songs.append(song)
+    dur = g.userhist['history'].duration
```
ids1024: This line doesn't seem to be doing anything.

kraetzin: My bad. Forgot to cut that out. I've removed it.

Issue #442, opened 2016-02-22
[Feature] Reverse playlist


flocke:
Like the title says I would love a command to quickly reverse an entire playlist.
A lot of playlists I load from Youtube are in inverse order and it is kind of tiresome to use `sw` to reverse them manually.

flocke:
Thanks for implementing it. I actually forgot that I submitted this request...
Just tested it and everything is working fine for me ;)

PR #473, opened 2016-04-11
Add a reverse command

kraetzin:
Referencing issue #442.

This implements the `reverse`, `reverse <number>-<number>`, and `reverse all` commands for editing playlists. `reverse all` is especially useful for #472 since it moves the latest played songs to the start of history.

### DIFF
--- a/mps_youtube/main.py
+++ b/mps_youtube/main.py
@@ -2641,6 +2641,12 @@ def reverse_songs_range(lower, upper):
 @commands.command(r'reverse all')
 def reverse_playlist():
     """ Reverse order of entire loaded playlist. """
+    # Prevent crash if no last query
+    if g.last_search_query == (None, None):
+        g.content = logo()
+        g.message = "No playlist loaded"
+        return
+
     songs_list_or_func = g.last_search_query[1]['func']
     if callable(songs_list_or_func):
         songs = reversed(songs_list_or_func(0,None))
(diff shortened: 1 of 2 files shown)


mps_youtube/main.py
```diff
@@ -2619,6 +2619,43 @@ def shuffle_fn():
...
+@commands.command(r'reverse all')
+def reverse_playlist():
+    """ Reverse order of entire loaded playlist. """
+    # Prevent crash if no last query
+    if g.last_search_query == (None, None):
+        g.content = logo()
+        g.message = "No playlist loaded"
+        return
+
+    songs_list_or_func = g.last_search_query[1]['func']
```
ids1024: This crashes with `KeyError: 'func'` for a playlist search (e.g. `//playlist`).

kraetzin: I hadn't considered playlist searches. I'll see if I can get those working.

kraetzin:
I've handled the error when trying to reverse a list of playlists. I also handled an error that occurs when trying to reverse the list of user uploads.

ids1024:
> I also handled an error that occurs when trying to reverse the list of user uploads.

I made this actually work in commit f439c1f6c80c3b98549f92235e0f7b2b0132f59e.

If `paginatesongs`, `IterSlicer` etc. seem somewhat convoluted, they sort of are; when  #423 is done the API for these things should be a lot cleaner.

PR #481, opened 2016-05-06
add youtube_dl as a dependency to the Dockerfile

PI-Victor:
*without this it will run into
https://github.com/mps-youtube/mps-youtube/issues/475 whenever trying to run mps
inside the docker container.


ids1024:
I wasn't merging this since youtube_dl is supposed to only be an optional dependency (though the  non-youtube-dl backend has a bug) but come to think of it, it is recomended, so the Dockerfile should probably install it.

PR #498, opened 2016-06-17
Make changing window title a configurable option

ardrabczyk:
I have plenty of these in my ~/.fluxbox/keys:

`Mod4 x         :exec wmctrl -a uxterm || uxterm`
`Mod4 e      :exec wmctrl -a emacs@ || emacs`
`(...)`

When I press Mod4-x fluxbox moves focus to an existing window titled
`uxterm' or it starts a new instance of uxterm if such window does not
exist.  When mpsyt changes its window title automatically it breaks
this workflow.  I have added a new config option called set_title
which enables/disables automatic changing of window title.  It is set
to True by default for backward compatibility.

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -305,7 +305,8 @@ class _Config(object):
             ConfigItem("audio_format", "auto",
                 allowed_values="auto webm m4a".split()),
             ConfigItem("api_key", "AIzaSyCIM4EzNqi1in22f4Z3Ru3iYvLaY8tc3bo",
-                check_fn=check_api_key)
+                check_fn=check_api_key),
+            ConfigItem("set_title", True)
             ] 
 
     def __getitem__(self, key):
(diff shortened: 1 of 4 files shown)


ids1024:
I think you should use `WM_CLASS` to identify uxterm instances, instead of the title. You can use the `xprop` command to find the value of this for a window, and it looks like it can be used in wmctrl with the `-x` argument.

So, try using `wmctrl -x -a UXTerm`.

ardrabczyk:
On Fri, Jun 17, 2016 at 05:46:36PM -0700, Ian D. Scott wrote:
> I think you should use WM_CLASS to identify uxterm instances, instead of the
> title. You can use the xprop command to find the value of this for a window,
> and it looks like it can be used in wmctrl with the -x argument.

Good idea, it works like charm. I have never thought about that,
thanks.

Now this patch is much less important but I think it's still valid.
Would you like to merge it or should I close it?

-- 
Arkadiusz Drabczyk <arkadiusz@drabczyk.org>

ids1024:
I think the PR should be closed, unless there is another reason to want this. Using `WM_CLASS` is the "correct" way to handle this. It is best to avoid adding too many configuration options to mpsyt.

PR #517, opened 2016-08-08
Added command --no-textart to disable textart, so users with screen reader wouldn't get confused by it

Gongreg:
Person at https://github.com/mps-youtube/mps-youtube/issues/515 asked for command to disable textart at start of mpsyt.


ids1024:
I wonder if there is a better way to deal with this. Ideally it could detect screen readers automatically. I don't know if that is possible...

Gongreg:
I am not sure about that too. On the other hand, if setting this command at startup helps for some people, thats the least we could do about it.

PR #524, opened 2016-08-26
Add mpv as a recommended player in mac installation section

sup:
Since mplayer is having issues playing certain tracks as mentioned in [Issue 518](https://github.com/mps-youtube/mps-youtube/issues/518), the README should be updated with mpv installation instructions for Mac OSX. 

I know I had issues when I installed mplayer first and got confused :P

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -99,7 +99,10 @@ Installing youtube_dl is highly recommended::
 
 Additional Mac OS X installation notes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-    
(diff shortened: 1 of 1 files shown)


ids1024:
Oh, the instructions still recommend mplayer? Mpv is definitely recommended now.

The `mplayer` instructions should probably be specifically labeled as an "alternative", and the windows instructions should be updated as well. I can do that myself though.

HobbinHood:
Should the brew formulae be updated to not solely depend on mplayer, but either mpv or mplayer instead?

PR #530, opened 2016-09-26
Allow exit code 3 as non-failed exit code

the-eater:
mpv now on default uses exit code 3 when exiting when done


ids1024:
That's odd... Is this documented anywhere? I don't see any mention in the release notes.

the-eater:
Wow I'm very sorry for wasting your time, I messed around with the `mpv` args and then set it to `''` which then it tried to play, and then it exited with exit code 3 because it couldn't play :') I am a horrible person. I wish you a good day

PR #533, opened 2016-10-02
Simple subscription system

jourd:
Added a simple subscription system. I did not know how to make the doc, i leave it to you. This is my first contribution to a free project so i apologize if i did anything wrong.

### DIFF
--- a/mps_youtube/commands/__init__.py
+++ b/mps_youtube/commands/__init__.py
@@ -23,5 +23,5 @@ def command(regex):
 
 
 # Placed at bottom to deal with cyclic imports
-from . import download, search, album_search, misc, config, local_playlist
+from . import download, search, album_search, misc, config, local_playlist, subscribe
 from . import play, songlist
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -8,7 +8,7 @@ import pafy
 
 from .. import g, c, screen, config, util, content
 from ..playlist import Video
-from . import command
+from . import command, subscribe
 from .songlist import plist, paginatesongs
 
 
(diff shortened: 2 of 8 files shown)


ids1024:
This seems like a good start, but not quite complete. For one thing, the `subs` command doesn't allow selecting a subscription to view its contents. Also, just having a count of the number of new videos is not really enough; there should be some way to see what specific videos are new (preferably also with a way to view new videos across all subscriptions).

alfaktulu:
I agree like Kodi's Youtube

silver-line:
To anyone who is still looking for mps-youtube to add support for getting your youtube subscriptions feed from your youtube account. As of right now that is basically not possible.

I wanted the feature so I was going to try and implement it, however after looking into the youtube API, they removed that ability a while back. With the current API you would have to retrieve a list of all the channels the user is subscribed to, then get the last few videos from every one of those channels, then sort them chronologically. This would be very slow and taxing, and would probably hit the quota very quickly.

The only other way is to just load the actual youtube sub page, and then parse the html. I may end up doing that for my setup, but I probably won't try to commit it here, as I think that goes against the youtube TOS.

I just wanted to put this here just incase anyone comes across this and wants to implement it. I can save them the trouble of finding out you can't. :/

Hopefully youtube will update the API at some point, but I wouldn't hold your breath because this has been an open issue for years on their API, and they have yet to respond. It would seem they are gimping the API on purpose in order to keep end users using their site and not a 3rd party app.

bsless:
Why would it be slow and taxing? Using the xml API seems like a viable option. You'll have to download as many xmls as the number of channels in the subscription feed, get all the entries from all the xmls, then sort them according to the 'published' attribute using `datetime.strptime` as key function.
Is it supposed to be costly? There's no need to constantly refresh the feed. You can add a command to manually refresh it for all channels or for a certain channel.
The xmlUrl for a channel is equivalent to the channel url:
`https://www.youtube.com/feeds/videos.xml?channel_id=<ID> <==> https://www.youtube.com/channel/<ID>`
So they can either be added through channel search or by downloading the subscription_manager.xml file from a user's subscription page.
I'd be happy to contribute / help around with that but I have little experience with contributing / developing large applications, but I'll be happy to learn as I go along. It's a feature I'd be really happy to see added.

hereforthebugs:
Couldn't you use the RSS feed generated on the "manage subscriptions" page? It gives you a .opml.

PR #548, opened 2016-10-20
Allow user to limit their searches to any, long, medium, and short du…

jas32096:
Allow user to limit their searches to any, long, medium, and short duration buckets.

With this PR, users can now use the syntax `/[<videoDuration>] <query>` to make searches with a given **videoDuration**.

`/[long] daft punk`


ids1024:
This seems like a good feature to add, and it seems to work properly; I'll have to think if this is the best syntax to use for it though.

PR #552, opened 2016-10-24
All allow user to search for videos/music after a certain time frame

jas32096:
All allow user to search for videos/music after a certain time frame

With this PR, users can now use the syntax `/<query> |<searchAfter>|` to search for videos after a certain time frame.

```
/daft punk |week|
/query |2016-10-17T00:00:00Z|
/query |2016-10-17|
```

### DIFF
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -3,6 +3,7 @@ import json
 import math
 import base64
 import logging
+from datetime import datetime, timedelta
 
 import pafy
 
@@ -14,10 +15,15 @@ from .songlist import plist, paginatesongs
 
 ISO8601_TIMEDUR_EX = re.compile(r'PT((\d{1,3})H)?((\d{1,3})M)?((\d{1,2})S)?')
 
+DAYS = dict(day = 1,
+            week = 7,
+            month = 30,
+            year = 365)
+
 
 def _search(progtext, qs=None, msg=None, failmsg=None):
     """ Perform memoized url fetch, display progtext. """
-    
+
     loadmsg = "Searching for '%s%s%s'" % (c.y, progtext, c.w)
 
     wdata = pafy.call_gdata('search', qs)
@@ -54,7 +60,7 @@ def token(page):
     return b64.strip('=')
 
 
-def generate_search_qs(term, match='term'):
+def generate_search_qs(term, match='term', after=None):
     """ Return query string. """
 
     aliases = dict(views='viewCount')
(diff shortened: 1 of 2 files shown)


ids1024:
As with your previous PR, the feature is fairly useful, but I'm not sure about the syntax. There are several issues:
- This could in theory break certain searches with the `|` character (`[` and `]` characters in #552). I don't think this is likely to be a major issue in practice, but it is worth considering.
- The syntax is not obvious, but I don't know if there is a way to make it so.
- It does not really leave too much room for other features of the sort, at least not in a consistent way.

A different syntax for these features like `[duration=long, after=2016-10-17]/query` or `/query --duration=long --after=2016-10-17` (I think I prefer the latter) would partially address these issues, but is not so concise. It could support a shorter form though, like `/query -d long -a 2016-10-17`.

What do you (or anyone else who has input) think of this? 

> I think 80 characters are too limiting. We should increase the line limit to 120, at least. There are already existing lines that exceed the 80char limit and most people are programming on widescreen monitors, not eighty character punchcards.

Fair point. An 80 character line limit is also useful for a two column layout. So you can have a text editor in one half of the screen and test the program in the other half. It's also a fairly standard rule; PEP8 recommends a 79 character limit: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

Thanks for the contribution anyway; designing good interfaces is just difficult.

jas32096:
I can see how the syntax could be confusing, but can't think of a better one. My regex skills are lacking, so I can't, easily, create a new command, with the existing decorator.

I did some testing on searches that include `|`, `[`, and `]` and haven't run into any issues. There must be some edge cases, but like you said, it will almost never be an issue. One thought, is that we could allow the user to escape it. I'm doing some checks on term-length and both the start and end, to minimize the chance of this becoming an issue.

Using *nix style flags, like `--duration` or `-d` still has the same issue of rare edge cases, but will increase the programming complexity. If you have a good way of parsing that, feel free to amend the PR. There must be something similar to bash's and C's getopt(s), for Python.

ids1024:
>  but will increase the programming complexity

Python has a module called `argparse` for parsing arguments; there is also a `getopt` module that works similarly to the c version.

I can implement that part. So do you agree that `--duration`/`-d` and `--after`/`-a` are good choices?

I'm not sure how much this may result in searches becoming invalid. It may be worth seeing if there's a way to get `argparse` or `getopt` to just ignore unrecognized arguments, which should largely eliminate that issue.

jas32096:
I'm not opposed to the `--duration/-d` and `--after/-a` syntax and agree it allows for more extensibility.

If possible, we could have extra arguments be passed to `generate_search_qs`, with `kwargs`. Then just do something like `qs.update(kwargs)`.

ids1024:
> If possible, we could have extra arguments be passed to generate_search_qs, with kwargs. Then just do something like qs.update(kwargs).

Sounds like a good idea. The way current code deals with this is to manipulate the return value of `generate_search_qs` after calling the function.

I'll have to think if there is a good way to integrate unix style arguments into mpsyt's command system. Currrently commands are parsed with regexs, so this would be part of a general improvement of them.

jas32096:
Alright; I did some playing around with argparse and came up with this

```python
from argparse import ArgumentParser

parser = ArgumentParser()
parser.add_argument('-d', '--duration', choices=('any', 'short', 'medium', 'long'))
parser.add_argument('-a', '--after')
parser.add_argument('search', nargs='+', type=str)

args = parser.parse_args('-d medium --after month this is the     rest of the    search'.split(' '))

for k,v in args._get_kwargs():
    print("%-10s %s" % (k + ':', v if k != 'search' else ' '.join(v)))
```

output
```
after:     month
duration:  medium
search:    this is the     rest of the    search
```

I could drop something like this in, in place of the current logic in **search**, but if you'd like to overhaul the command system, I recommend you take it from here.

ids1024:
Yeah, that should do for now. Redesigning the command system is a good long term goal, but it is going to be complicated (it will likely require breaking changes to the syntax of some commands).

jas32096:
I made the changes in #553; seems to be working.

PR #553, opened 2016-10-24
Merge #548 and #552

jas32096:
Closes #548 and #552

### DIFF
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -3,6 +3,7 @@ import json
 import math
 import base64
 import logging
+from datetime import datetime, timedelta
 
 import pafy
 
@@ -13,6 +14,11 @@ from .songlist import plist, paginatesongs
 
 
 ISO8601_TIMEDUR_EX = re.compile(r'PT((\d{1,3})H)?((\d{1,3})M)?((\d{1,2})S)?')
+
+DAYS = dict(day = 1,
+            week = 7,
+            month = 30,
+            year = 365)
 
 
 def _search(progtext, qs=None, msg=None, failmsg=None):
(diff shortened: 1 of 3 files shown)


jas32096:
Alright. Should be working now, with the new syntax.

Note: For invoking from the command line, you'll need to surround the search with quotes like:
&nbsp;&nbsp;&nbsp;&nbsp;`mpsyt '/best music -d long -a year'`

ids1024:
I found an issue with this. If you try a command like `/query --some-invalid-argument` or `/query --help`, it prints a message then exits. The message is fine, but it shouldn't terminate mpsyt.

I don't know if there is a way to change that behavior with `argparse`. I don't see anything in the docs.

jas32096:
Right. I also forgot about passing on extra args with, kwargs, but I'm not gunna do that, for the time being. I'll try surrounding the parse with a try/except

jas32096:
Made some changes and really tried to beat it up this time.

One of the last things I tested with was `/ --this -qwerty -a 2015-09-22 this is a test --duration long --bad -c --`, which is was able to handle fine, ignoring the unused arguments.

mps_youtube/commands/search.py
```diff
@@ -218,10 +218,14 @@ def related_search(vitem):
...
-    args = parser.parse_args(term.split(' '))
-    videoDuration = args.duration if args.duration else 'any'
-    after = args.after
-    term = ' '.join(args.search)
+    try:     #TODO make use of unkowns
+        args, unknown = parser.parse_known_args(term.split(' '))
+        videoDuration = args.duration if args.duration else 'any'
+        after = args.after
+        term = ' '.join(args.search)
+    except SystemExit as e:  #<------ argsparse calls exit()
```
ids1024: Oh, I didn't know that `sys.exit()` triggers a catchable exception. Nice!

PR #574, opened 2017-01-05
player.py: use TLS when retrieving art

vszakats:


ids1024:
Pull requests should be made to the `develop` branch. I've cherry-picked the commit there.

That aside, thanks! This is obviously not a major security issue, but I agree that it is generally best to use https.

PR #619, opened 2017-04-15
Fix misuse of flags in re.sub() call

jwilk:
The 4th argument of `re.sub()` is maximum number of substitutions, not flags.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


ids1024:
It looks like you're right. Thanks for pointing this out!

PR #621, opened 2017-04-27
Feature request: allow user to list yt video ID in search and playlist results. 

buzzbar:
Feature is simple, can be implemented with a configuration parameter, and would provide a column for the user in the search or playlist results indicating the YouTube video id, much like how it's displayed while streaming. 

This feature would be very handy for those inclined to command and control mpyst via a shell script and the like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ parts
 bin
 var
 tags
+.env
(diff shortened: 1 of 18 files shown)


ids1024:
This should be an issue, not a pull request.

Issue #618, opened 2017-04-14
USER SEARCH: [user bbc] interpreted as /user/bbcnews/videos, and not /user/bbc/videos


T-jw:
Not sure if this is user error or a bug, but I would expect searching '**user BBC**' to give me the results for '**BBC**' before '**BBC News**'. Is there a workaround?

Thanks!

![screenshot from 2017-04-15 05-45-58](https://cloud.githubusercontent.com/assets/5573276/25051215/9dc89bb8-219f-11e7-81cf-9b1cb0b9d343.png)
Screenshot showing webpage: /user/bbc/videos, and webpage: /user/bbcnews/videos, and mpsyt resutls for: user BBC.

ids1024:
To get the user, mpsyt runs a channel search, and selects the first result. This behavior is clearly not desirable, so perhaps there is a better way.

Razesdark:
I'm pretty new to Python so I'm not sure if this code is something I should submit, but I added some SequenceMatcher stuff to channelfromname, so it looks for the channel name most similar to 
the search query.
```python
                ba = {}
                bal = 0
                for alt in userinfo:
                    # Get the title
                    au = alt.get('snippet', {}).get('title')

                    # Compare and select channel with name most
                    # similar to search query
                    al = SequenceMatcher(None, au, user).ratio()
                    if (al > bal):
                        ba, bal = alt, al

                del bal
                snippet = ba.get('snippet', {})
```

I've tried to search for some random channels, `user BBC` now fetches the correct BBC channel.

ids1024:
@Razesdark That sounds better than the current behavior. But I'm not sure it's ideal.

I think what would be best is for the command to only work with exact matches (if the API has a way to do that), and have a separate command that implements a user search, somewhat like the current playlist search.

The second part of that is non-trivial to implement, but probably not unreasonably difficult.

Razesdark:
Using the `youtube.channels.list` endpoint instead of the `search.list` endpoint you can find a list of channels based on a username. Using that you get a case insensitive, non-fuzzy search.

The returned JSON for the endpoint suggests that one username may hold more than one channel, but I have not found any that actually has more than one. When testing it on some small and some large YT channels it has only given me the expected results. 

I'm opening a PR for that. Then I'll work on a `usersearch` functionality. It might take some time, I'm still new to OS and python.

ids1024:
> The returned JSON for the endpoint suggests that one username may hold more than one channel...

[...]

PR #628, opened 2017-05-02
Improves user search per #618

Razesdark:
Switches from search.list to channel.list api endpoint for user search.

### DIFF
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -13,7 +13,7 @@ parser.add_argument('search', nargs='+')
 
 import pafy
 
-from .. import g, c, screen, config, util, content
+from .. import g, c, screen, config, util, content, listview, contentquery
 from ..playlist import Video
 from . import command
 from .songlist import plist, paginatesongs
@@ -150,6 +150,32 @@ def channelfromname(user):
     return (user, channel_id)
 
 
+@command(r'channels\s+(.+)')
+def channelsearch(q_user):
+
+    qs = {'part': 'id,snippet',
+          'q': q_user,
+          'maxResults': 50,
+          'type': 'channel',
+          'order': "relevance"
+          }
+
+    QueryObj = contentquery.ContentQuery(listview.ListUser, 'search', qs)
+    columns = [
+        {"name": "idx", "size": 3, "heading": "Num"},
+        {"name": "name", "size": 30, "heading": "Username"},
+        {"name": "description", "size": "remaining", "heading": "Description"},
+        ]
+
+    def run_m(user_id):
+        """ Search ! """
+        usersearch_id(*user_id)
+    del g.content
+
+    g.content = listview.ListView(columns, QueryObj, run_m)
+    g.message = "Results for channel search: '%s'" % q_user
+
+
 @command(r'user\s+(.+)')
 def usersearch(q_user, identify='forUsername'):
     """ Fetch uploads by a YouTube user. """
@@ -157,7 +183,7 @@ def usersearch(q_user, identify='forUsername'):
     user, _, term = (x.strip() for x in q_user.partition("/"))
     if identify == 'forUsername':
         ret = channelfromname(user)
-        if not ret: # Error
+        if not ret:  # Error
             return
         user, channel_id = ret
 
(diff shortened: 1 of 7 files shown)


ids1024:
Yes, that's the way to do it, and it seems to work properly. I think it's probably best not to merge until searching is implemented (which I presume is your intent in tagging it WIP) since it breaks imprecise user searches.

Razesdark:
Ok, I've been making some potential progress this week. I have spent some time making a content agnostic listview. The code is stilll pretty hacked together, and not very delicate so I won't upload until its a bit cleaner. But since I can't work on this for a week or so, I'll write it up so I can get some feedback on it.

Here is the current user search view. (Plus some debugstrings i forgot to clean up).
test

I've been trying to work on a content agnostic list view and a few supporting functions and classes, reusing a lot of the generate_playlist_code. It differs a little bit from how PaginatedContent was expected to work as this view remembers its own page number. I also made it so that a function named _run() can be added to any content.Content derived classes which gets priority when matching userinput, so future uses of content.Content can hold its own functions. I use this to change pages. When you create a view, you can specify which function it is going to run when you've selected your candidate from the list.

I also made a wrapper class for call_gdata, that emulates a list of all search results, but it will only query for more pages when needed.

ids1024:
That generally sounds good, though I can't comment on the exact details without seeing the code.

The content API is a work in progress, and certainly not perfect, so any improvement is welcome. Originally `g.content` contained just strings. In #423 I tried to get the content API to be used for everything, but that stalled because it was somewhat complicated and I've been working on other projects.

Razesdark:
Alright, I can take a look into that commit over the next few weeks and the code and see what I can do about it. I see some of those points will end up being covered by ListView, potentially.

ids1024:
> So a single content class that we build subclasses for each kind of view from, making a unified way of rendering and interfacing with those derived classes so we have all code relevant to a single view in a single location would be a OK way to move forward?

That sounds good. Keep in mind that we want to move to a curses-based interface (which is why I began refactoring with a `Content` class a while ago), so it needs to be compatible with that. I think this should work for that (perhaps the `.content()` method could be done differently with a curses ui, but the important thing is that the relevant code is all in one place.)

mps_youtube/commands/search.py
```diff
@@ -151,14 +150,40 @@ def channelfromname(user):
...
+    columns = [
+        {"name": "idx", "size": 3, "heading": "Num"},
+        {"name": "name", "size": 30, "heading": "Username"},
+        {"name": "description", "size": "remaining", "heading": "Description"},
+        ]
+
+    def run_m(user_id):
+        """ Search ! """
+        usersearch_id(*user_id)
+    del g.content
```
ids1024: Why would this be necessary?

mps_youtube/listview.py
```diff
@@ -0,0 +1,237 @@
+"""
+    DOCSTING COMES HERE
+"""
+import re
+
+from . import c, util, content
+
+
+class ListViewItem(object):
```
ids1024: Mpsyt is now Python 3 only, so explicitly inheriting from `object` is redundant.

mps_youtube/commands/search.py
Razesdark: It isn't. I found a bug in the code where a second channel search would still yield the results from the first search. I wasn't entirely sure what caused this, so I added some code to make sure i deleted previous data before rewriting them, . When I found the bug, i removed much of that code, seems I skipped this line.

mps_youtube/listview.py
```diff
@@ -0,0 +1,237 @@
...
+        types = len(set([obj.__class__ for obj in objects]))
+        if types == 0:
+            raise BaseException("No objects in list")
+        if types > 1:
+            raise BaseException("More than one kind of objects in list")
+
+        self.object_type = [obj.__class__ for obj in objects][0]
+        del types
+
+    def pages(self):
```
ids1024: This function is just the equivalent of `max(1, len(self.objects) / self.views_per_page())`, right?

ids1024: Plus a call to `math.ceil`

Razesdark: Yea, ceil was was i was going for.

ids1024:
Merged with some changes; thanks! I made ListView a subclass of `PaginatedContent` (8779160ae185680bc80cc35e0a7b02d19559c86c). I agree that `g.current_page` should be part of the content API rather than a global, but that can be dealt with when a content API based solution is used for playlist search and `paginatesongs()`, so there is no duplication.

Issue #634, opened 2017-05-17
[feature-request] add command to list live stream in youtube and support them?


uNouss:
Hello thank you for this great software, I use it to the everyday and I am very satisfied with it. So thank you for the big work. I would want just to propose a feature which would be cool to add. On YouTube, we can see for example here https://www.youtube.com/channel/UC4R8DWoMoI7CAwX8_LjQHig , live videos. So I would love one command `live` for example which allows to list them and being able to read them directly since mpsyt or maybe something like `live -c news|sport|whatever` this to list them by categories.

Razesdark:
Hi! I think a live stream search is a great idea, I've worked a bit on it and got a working prototype ready, using a category search and limit the results to livestreams.

Just to make sure, you are able to watch livestreams with the current version of the software, this feature-request is only about the search function?

uNouss:
Only the search functionality is required with the category functions. The livestream is already working on the example I tested.
![2 _041](https://cloud.githubusercontent.com/assets/15982321/26233561/8e24340e-3c5e-11e7-86a6-704c59c5e765.png)
![2 _040](https://cloud.githubusercontent.com/assets/15982321/26233562/8ec647c6-3c5e-11e7-8593-7f8a43689cb9.png)

Razesdark:
Thanks for the feedback.

This functionality has been merged into the develop branch(9e3006e) now.
In case you haven't read the pull request the gist of it is that you can now use the `--live` and `--category` when searching in order to search for live streams.

You can also view entire categories by using the command `live <category>`, to search entire categories. We have only added support for some of the most normal categories. see `help search` for more information

uNouss:
Hello thank you for the functionality, I installed the development version, but it does not work as expected. The basic mpsyt commands do not work (example: when I try `live something` and do: `all`, or `all -a` or `1` or `1 -a` ...) it does not play live). https://asciinema.org/a/a67p5jlc0cjs27s2iuzi8cn24 Here's a catch to see what's going on. (Anything else, would it be possible to filter in the global search - not just for livestreams - a keyword or user filtering option? like this `/search -f user|word`) 
Anyway thanks for the quick work.

Razesdark:
Hi! 
I will push a new update that fixes the `--category` bug. As far the -a, parameter, it totally slipped off my radar. I'll look into that as well.

If you want to fix this before the push, this change will fix the --category issue.

Line `293` in `mps_youtube/commands/search.py,` that looks like
[...]

PR #635, opened 2017-05-18
Implements livestream search #634

Razesdark:
Implements a simple livestream category search using the "live" command as described in the issue #634.

I reckon we could add a livestream filter to the search rather than have live take more arguments

### DIFF
--- a/mps_youtube/g.py
+++ b/mps_youtube/g.py
@@ -51,6 +51,16 @@ PLFILE = os.path.join(paths.get_config_dir(), "playlist_v2")
 HISTFILE = os.path.join(paths.get_config_dir(), "play_history")
 CACHEFILE = os.path.join(paths.get_config_dir(), "cache_py_" + sys.version[0:5])
 READLINE_FILE = None
+categories = {
+        "film":      1,
+        "autos":     2,
+        "music":    10,
+        "sports":   17,
+        "travel":   19,
+        "gaming":   20,
+        "blogging": 21,
+        "news":     25
+}
 playerargs_defaults = {
     "mpv": {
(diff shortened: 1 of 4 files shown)


mps_youtube/g.py
```diff
@@ -51,6 +51,16 @@
 HISTFILE = os.path.join(paths.get_config_dir(), "play_history")
 CACHEFILE = os.path.join(paths.get_config_dir(), "cache_py_" + sys.version[0:5])
 READLINE_FILE = None
+categories = {
```
ids1024: It looks like the categories can be retrieved through the API: https://developers.google.com/youtube/v3/docs/videoCategories/list

It might not be great for this purpose though, since the titles are several words.

ids1024:
Perhaps instead of a separate command, a syntax for the search command like `search -live -c <category>` could be added?

mps_youtube/g.py
Razesdark: The names were a big reason why I went with a simpler solution like this. 

Also, since that same list is used to classify normal videos, many of the categories may not be that applicable to live streams so we would clutter the output with a lot of potentially empty categories(since we're not polling to check that they have active channels)

Razesdark:
Adding more arguments to get more fine grain control of searches is good, however there is something very simple about being able to write `live music` or `live news` and get a list of applicable channels. 

There is some overlap between a full category search and searching within a category but I think facilitating both could be worth.

ids1024:
Fair point; it is more concise.

ids1024:
Help text for this command should be added to `helptext.py`.

Issue #610, opened 2017-03-30
Refactor player.py to be class based


ids1024:
There should be a `Player` class, with `Mpv` and `Mplayer` sub-classes. This is cleaner and should help allow the addition of new back-ends, like  mpd (#168) or chromecast (#398 #607).

georgeaueb13:
Hello, I am George Metaxopoulos!We had contacted you via email recently and you told us that we could do the refactoring as part of our semester project with another student(@panosgt7)Do you mind assigning it to us?
@ids1024

catafest:
can you help me with this ?

https://github.com/mps-youtube/mps-youtube/issues/611

2017-03-31 14:34 GMT+03:00 georgeaueb13 <notifications@github.com>:

ids1024:
@georgeaueb13 @panosgt7 It seems GitHub doesn't allow assigning an issue to someone who isn't a collaborator / member of the organization (which is annoying), but consider yourselves assigned.

ata2001:
@georgeaueb13 Hi! Are you still working on this?

georgeaueb13:
@ata2001 Hello! We got caught up with some university work the past few weeks. Starting tomorrow we will keep on working on the project and we hope to have everything done by the end of the month!

ata2001:
@georgeaueb13 Thanks for the reply! Good luck!

vn-ki:
This is closed in #753

PR #640, opened 2017-06-06
Refactor player.py to be class based #610

gmetaxo:
This is our approach(along with @panosgt7) to making player.py class based. Player is an abstract base class and mplayer and mpv are subclasses of this class. Right now, Player simply serves as a model of how we think the project should continue builiding. Generic methods for all players should be added in Player class(instead of the ones we have as example right now), in order to provide some sort of documentation-layout for the addition of new players. The subclasses now separately implement the already created methods(with some needed changes) for each player. We hope that you agree with this design and that we can all continue improving our approach until it is ready for integration.

P.S. 1:Setting up the project with setup.py did not work after your last additions to the project, so maybe there are problems with building the project with our work. We have also implemented it on top of the latest stable version and it seems to work in the same manner as it did before we applied them.

P.S. 2:This is our first time contributing to an open source project, we hope that we have built something we could all work on and improve. Any comments are welcome.


mps_youtube/commands/play.py
```diff
@@ -92,7 +84,21 @@ def play(pre, choice, post=""):
...
+                return
+            else:
+                if config.PLAYER.get.endswith("mpv.exe"):
+                    mpv = Mpv()
+                    mpv.play_range(songlist, shuffle, repeat, override)
+                elif config.PLAYER.get.endswith("mplayer.exe"):
+                    mplayer = Mplayer()
+                    mplayer.play_range(songlist, shuffle, repeat, override)
+                else:
+                    g.message = "Player not supported! Enter %sset player <player_app> "\
```
ids1024: Mpsyt currently does "support" alternate players (vlc for instance should work, although it is not integrated); it should have a fallback that just calls the player with the URL.

mps_youtube/commands/play.py
```diff
@@ -7,8 +7,8 @@
 from . import command, WORD, RS
 from .songlist import plist
 from .search import yt_url, related
-from ..player import play_range
-
+from ..players.mpv import Mpv
+from ..players.mplayer import Mplayer
```
Razesdark: Shouldn't we somehow load only the player(s) we're actually using?

mps_youtube/commands/play.py
```diff
@@ -92,7 +84,21 @@ def play(pre, choice, post=""):
                 streams.preload(g.model[chosen + 1], override=override)
 
         try:
-            play_range(songlist, shuffle, repeat, override)
+            if not config.PLAYER.get or not util.has_exefile(config.PLAYER.get):
+                g.message = "Player not configured! Enter %sset player <player_app> "\
+                    "%s to set a player" % (c.g, c.w)
+                return
+            else:
+                if config.PLAYER.get.endswith("mpv.exe"):
```
Razesdark: using .exe in the endswith() calls breaks playback on linux.

We could change these if statements to if "mpv"/"mplayer" in config.PLAYER.get and it should work for most people unless you put your mpv binary in a folder called mplayer or vice versa.

mps_youtube/commands/search.py
```diff
@@ -271,12 +273,28 @@ def livestream_category_search(term):
...
     def start_stream(returned):
         songs = Playlist("Search Results", [Video(*x) for x in returned])
         player.play_range(songs, False, False, False)
 
+        if not config.PLAYER.get or not util.has_exefile(config.PLAYER.get):
+            g.message = "Player not configured! Enter %sset player <player_app> "\
+                "%s to set a player" % (c.g, c.w)
+            return
+        else:
+            if config.PLAYER.get.endswith("mpv.exe"):
```
Razesdark: Same as I mentioned before.

mps_youtube/commands/play.py
ids1024: That shouldn't really be a problem, unless there are so many players in a future version that it substantially slows down startup (unlikely).

mps_youtube/commands/play.py
georgeaueb13: We will fix that with @panosgt7  and once it is ready we will update our pull request!

mps_youtube/commands/play.py
georgeaueb13: We will apply a fix for that as well and once it is ready we can discuss it further.

mps_youtube/commands/play.py
georgeaueb13: We are going to need some help on that. As you said, vlc is not fully integrated. Since vlc was using play_range in order to work, perhaps we can create another subclass for vlc, similar to the mpv and mplayer ones(with proper additions/deletions), and then improvements can be integrated there if someone wants to do that. Still, this would only work for vlc, if another player is used(other than the 3 "supported" ones) then the "else" part should remain the same. So let me ask:
1)Do you agree with this approach or is there a different way you would prefer? 
2)Are other players partially supported? Because, other than mpv and mplayer which are fully supported, and vlc, which is partially supported, we have not yet encountered such a case. If so, is this something you want? Shouldn't we try and integrate as many players as possible via subclasses instead of relying on the current program to work under `set player unintegrated_player_name` ?

ids1024: It's probably best just to have a generic fallback, although a dedicated `Vlc` class and such would be good to have in the long run, since I don't know what players people might be using and we want to try not to break them.

Mpd and catt for example are known to work (https://github.com/mps-youtube/mps-youtube/issues/168, https://github.com/mps-youtube/mps-youtube/issues/607), although not perfectly.

georgeaueb13: Maybe we could do both. Provide a specific class for vlc and a general fallback for any other player. We will look at the code more precisely in order to see all the methods etc. needed for the other players. Would you prefer that in a different file, a new subclass for all other players or a general method-methods in Player class? I believe the second option is the best.

ids1024: Sounds good. After all, part of the point of refactoring this code is to be able to support more players properly.

georgeaueb13:
I made most of the suggested changes.There are some import statements that can be omitted, I will run Pylint later to fix them quickly.
I added a check_player method in Player class that checks which player is set and return the proper subclass  object. This method is now called in play.py and search.py. I also created "Vlc" and "OtherPlayers" classes that I think keep the same funcionality as the current version(maybe some more statements can be omitted from their methods, but I wasn't 100% sure so I kept some of them that don't seem to raise exceptions). I look forward to your review.

mps_youtube/players/other_players.py
```diff
@@ -0,0 +1,254 @@
...
+                logging.info("Keyboard Interrupt")
+                util.xprint(c.w + "Stopping...                          ")
+                screen.reset_terminal()
+                g.message = c.y + "Playback halted" + c.w
+                raise KeyboardInterrupt
+                break
+            if config.SET_TITLE.get:
+                util.set_window_title("mpsyt")
+
+            if returncode == 42:
```
ids1024: These two magic return codes only apply to mpv/mplayer (using the special configuration mpsyt provides).

ids1024:
This is adding a fair amount of duplicated code; that should be avoided where possible. Probably some of the duplicated code could be moved into methods of the base `Player` class, calling out to simpler player-specific methods.

georgeaueb13:
Yes, we noticed that as well. Yet, it would require many changes in order to finally come to a design that would support this concept, as player.py was very complicated and dividing it into more general method would need much time. Right now it is exam period and the next two weeks are really busy. Maybe someone else or even you could help on this, otherwise we would have to continue with it after the exams. Finally, are other players supported by this code or it would require changes on that part as well?

ids1024:
I think it's probably best not to merge this until the duplication is addressed, but you are quite right that it will require a fair amount of work. I don't personally have the time to work on it at the moment, but someone else can if they want to.

georgeaueb13:
I will take a look on it tonight, mostly to evaluate what changes are needed and provide a feedback for any contributor that might be interested until we can get back to serious work with it. If you or someone else can tell us if other players are properly supported right now(not integrated, just partially supported as before), that would be really helpful.

PR #642, opened 2017-06-07
Added genpl command to automatically generate a playlist from text

nishanthkarthik:
## genpl

The `genpl` command takes a text file (containing song titles), finds the songs and creates a playlist out of them. This is how it works,

### Original text file - `songs.txt`

less

### Searching the best match for each title

inprocess

### Playlist created with a random name

done

### Playlist contents

result

This is my second contribution to this project. Please do suggest corrections/improvements 😃

### DIFF
--- a/mps_youtube/commands/__init__.py
+++ b/mps_youtube/commands/__init__.py
@@ -24,4 +24,4 @@ def command(regex):
 
 # Placed at bottom to deal with cyclic imports
 from . import download, search, album_search, misc, config, local_playlist
-from . import play, songlist
+from . import play, songlist, generate_playlist
--- a/mps_youtube/commands/album_search.py
+++ b/mps_youtube/commands/album_search.py
@@ -49,10 +49,11 @@ def _do_query(url, query, err='query failed', report=False):
     return wdata if not report else (wdata, False)
 
 
-def _best_song_match(songs, title, duration):
+def _best_song_match(songs, title, duration, titleweight, durationweight):
     """ Select best matching song based on title, length.
 
-    Score from 0 to 1 where 1 is best.
+    Score from 0 to 1 where 1 is best. titleweight and durationweight
+    parameters added to enable function usage when duration can't be guessed
 
     """
     # pylint: disable=R0914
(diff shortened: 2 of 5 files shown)


ids1024:
This looks potentially useful, but I wonder if another name would be clearer... I'm not sure what.

Razesdark:
Maybe we, at the same time could have it support output from video descriptions and comments(to make playlists from mix videoes and similar)

nishanthkarthik:
@ids1024 How about `mkp`? It is as unix-y as `rmp` or `vp`.

@Razesdark It is a good idea to look beyond titles. Currently the implementation does a single search, takes the topmost result and assumes it to be the best match. However, adding video description as a search parameter is negative in a way, since a lot of videos add credit to the song (that they use as a background music for their video) in the description text. This might create false positives.

Razesdark:
I think you misunderstood, take for example mix vidoes which have several songs in them. I am suggesting we use a videos description or comments as input instead of just files.

nishanthkarthik:
Oh. My bad. I get it now.

Besides, do you have any idea for the command name?

On Jun 8, 2017 4:03 PM, Tommy Stigen Olsen <notifications@github.com> wrote:

I think you misunderstood, take for example mix vidoes which have several songs in them. I am suggesting we use a videos description or comments as input instead of just files.

ids1024:
The matching should probably be done using `_best_song_match()` in `mps_youtube/commands/album_search.py`.

Or call `_match_tracks()` which calls that function, but it probably would need to be modified to allow using without an album title.

Razesdark:
Ok, so I've been trying to teach myself some regex and to see if i could make something that would extract songlike strings from comments and descriptions.

Its just some test code, but it does extract all non malformed song/artist combos from some random mixes I listen to. 

https://github.com/Razesdark/mps-youtube/blob/tracklist_test_code/mps_youtube/filename_extractor.py

I don't know if I modify the code in this PR, if that code can be applied to this PR.

nishanthkarthik:
@Razesdark Shall we have it as a separate issue/PR? It would be better if we had a full-fledged discussion on how to generalize these regex-es for any youtube video description.

For a different project some time back, I tried getting song metadata(artist/album) through regex and found it hard to use the same code for multiple videos.

@ids1024 Would you like any further changes on the PR?

PR #649, opened 2017-06-14
Generate playlists from songs listed in video descriptions #642

Razesdark:
As discussed in PR #642, I've extended the mkp function to allow users to make playlists from text found in video descriptions.

I have had the results pop up in a listview, so the user can inspect the parse before committing it to a playlist. Like any playlist selection, the user can select using numbers and ranges which songs they want to include in their playlist.

Finally I did some minor pylinting on the generate_playlist file.

record

### DIFF
--- a/mps_youtube/commands/generate_playlist.py
+++ b/mps_youtube/commands/generate_playlist.py
@@ -3,7 +3,7 @@ import pafy
 from random import choice
 import string
 
-from .. import content, g, playlists, screen, util
+from .. import content, g, playlists, screen, util, listview
 from ..playlist import Playlist
 from . import command, search, album_search
 
@@ -11,6 +11,12 @@ from . import command, search, album_search
 @command(r'mkp\s*(.{1,100})')
 def generate_playlist(sourcefile):
     """Generate a playlist from video titles in sourcefile"""
+
+    # Hooks into this, check if the argument --description is present
+    if "--description" in sourcefile:
+        description_generator(sourcefile)
+        return
+
     expanded_sourcefile = path.expanduser(sourcefile)
     if not check_sourcefile(expanded_sourcefile):
         g.message = util.F('mkp empty') % expanded_sourcefile
@@ -46,7 +52,7 @@ def create_playlist(queries, title=None):
     Create playlist with a random name, get the first
     match for each title in queries and append it to the playlist
     """
-    plname = title or random_plname()
+    plname = title.replace(" ", "-") or random_plname()
     if not g.userpl.get(plname):
         g.userpl[plname] = Playlist(plname)
     for query in queries:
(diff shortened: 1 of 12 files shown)


ids1024:
Videos with track lists in the description generally have all those songs in them, right? So the point here is to find alternate versions that are individual tracks, so you can treat them as individual tracks?

rocketinventor:
Documentation?

ids1024:
This seems like a fairly useful feature in certain cases; making it work reliably is difficult though. It seems to result in nonsense for videos that don't have track listings in their descriptions; perhaps that is acceptable, but it might be better to look for certain patterns and only match if it seems to have a track listing. That could potentially reduce the number of cases it matches successfully on, though.

It also isn't working too consistently; videos like https://youtube.com/watch?v=_-sYj_1lADo don't match correctly even though there is a track listing. Ultimately there's a limit to how reliable it can be without hard AI though.

ids1024:
One thing to do is to check for lines starting with times; those are a good indicator for a track listing.

mps_youtube/commands/generate_playlist.py
```diff
@@ -76,3 +85,43 @@ def random_plname():
...
+    ]
+
+    def run_m(idx):
+        """ Create playlist based on the
+            results selected
+        """
+        create_playlist(idx, title)
+
+    data = [listview.ListSongtitle(x) for x in data]
+    g.content = listview.ListView(columns, data, run_m)
```
Razesdark: Yea, I guess if the software is unable to find any tracks, it would throw exceptions. I'll add a check for that.

Razesdark:
@ids1024 Yeah, correct. My personal use case was with EDM or other mixes where songs are often mashed together and/or mixed together to a coherent long mix, so if I find songs I like in a mix and want to listen to them with their proper start and end, I need to search for them separately.

I didn't include straight up album searches, because we can already do that with the `album` command. Currently I use dashes to look for `artist` - `track title` like strings and use them, but i realize that does leave it unable to recognize single band mixtapes and such where the timestamp can be ommited and you only list titles. I am stumped on how to make it reliable. I guess most mixes do include timestamps, but I found a fair few that don't. Thumbs up for Rush!

@rocketinventor Sorry, I'm new to all of this, anything in particular i should document?

ids1024:
Yeah, albums are not the most useful example since `album` exists, but that was the easiest thing to test it with.

I don't know how reliable this can be made; it is expected that it will no be entirely reliable. But some things to try:
- If there is a series of lines starting with times, take that as the track listing and use those with the times stripped (possibly using the times to match, though those usually are the time in the video, so they would need to be subtracted to get durations)
- Try to support lists without dashes; a playlist of tracks by a single artist won't have the artist name
- Any other tricks

rocketinventor:
For the regex, first check for (0-1) matches that look like a timestamp, than match for the song info after that. It might be a good idea to tie into some sort of music database and search that to figure out the exact song metadata. There is no guarantee that song info will be in any specific order or format.

rocketinventor:
@Razesdark I would add something to the Readme explaining that this feature exists and explain the steps that need to be taken to use this feature (i.e. the commands that you typed in the gif that you shared).

rocketinventor:
I noticed that there is some explanation of this feature at https://github.com/mps-youtube/mps-youtube/pull/642 but there is no mention of this feature in the Readme. New users will likely be unaware of this feature.

Razesdark:
Ok, here is a updated version of the script. It tries to check each line and come up with a confidence index wether or not the line is a song title or not. 
Its based on a few factors, like does the line start with a number, does it look like a artist - track combination or does it have a time stamp.

Then it takes the lines with an index above average, which I feel works because tracklisting formatting is usually consistent so all lines should be equally above average, as long as there is one line or more weighing down the average. 

Once it has the list that it percieves as song titles, it tries to determine if the line has the artist in it or not. Checks wether or not the results are consistent, removes inconsistent results if they are few, or exits if it finds too many of them.

If it needs to find the artist name, it tries to parse the title by doing a youtube search on the title, then tries to repeatedly find the most frequent common string between a random subset of the results. Most of the time I tried this, it would only return the correct name, but sometimes it would have several results. (Like Best of, Album titles, etc), in the case of several alternatives, it will use the one that matches first in the original title. As I expect that the most common setup would be Artist name first.

rocketinventor:
@Razesdark Nice Work!

Issue #659, opened 2017-07-21
Save player options when using repeat mode


ritiek:
Not sure if this is possible either but at the moment one has to manually set the preferred playback speed and volume every time the track repeats itself. It would be nice if mps-youtube took care of it automatically.

SirWojtek:
Agree. Please notice that options also return to a default value when playing a playlist. This could be frustrating when you have to volume down in each new song :wink:

PR #660, opened 2017-07-24
Preserve mpv-player options and cache when using repeat mode

### LINKED ISSUE #658: Save video/track temporarily when using repeat
I am not sure if this is possible but It would be nice if mps-youtube could save the video/track temporarily side-by-side when using repeat mode, so it could just read from the saved file every time track starts over again. The track otherwise has to be downloaded again every time it starts over.

ritiek:
This feature only come in action when using `repeat mode` with `mpv-player` and `len(songlist) == 1`. I tried something like this with `mplayer` too but it resets everything whenever the song starts over again. So, haven't messed up anything with `mplayer`. This PR modifies and works only for `mpv-player`.

Partially fix #658 and #659.

### DIFF
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -38,8 +38,11 @@ def play_range(songlist, shuffle=False, repeat=False, override=False):
 
         if config.SET_TITLE.get:
             util.set_window_title(song.title + " - mpsyt")
+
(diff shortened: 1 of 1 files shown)


ids1024:
Thanks!

PR #664, opened 2017-08-03
Fixes #663

### LINKED ISSUE #663: mpsyt crash
Since today, mpsyt is crashing whenever I'm  searching songs or even playlists for DJ KSHMR

Steps to reproduce:

1. open mpsyt
2. search for 'KSHMR' by typing `/KSHMR` or `//KSHMR`

ritiek:
Fixes #663.

The problem was `mpsyt` attempting to parse a channel result using `get_track_id_from_json(item)` whereas the function expects a video result as parameter. I changed the code a bit such that only video results are passed into the function and anything else gets ignored.

P.S. Sorry for the unnecessary commits, you should probably squash them.

### DIFF
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -307,7 +307,7 @@ def search(term):
     logging.info("search for %s", term)
     query = generate_search_qs(term, videoDuration=video_duration, after=after,
                                category=args.category, is_live=args.live)
-    
+
     msg = "Search results for %s%s%s" % (c.y, term, c.w)
     failmsg = "Found nothing for %s%s%s" % (c.y, term, c.w)
     _search(term, query, msg, failmsg)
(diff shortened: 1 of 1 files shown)


modalduality:
Is ignoring the best thing to do? What is the proper way to show playlists? By the way, you can have filters inside the list comprehension:

`[get_track_id_from_json(i) for i in items if i['id']['kind'] == 'youtube#video']`

ritiek:
The playlist searching had the same problem as track searching. Fixed it and switched to list comprehensions.

I also tried using YouTube inbuilt search filters (for only video or playlist results) but they do not seem to work. Not sure if ignoring them is the best idea but I could not come with anything better.

ids1024:
It seems strange that channels would be mixed in to the results for video searches and playlist searches; the API query should prevent that. But if that is indeed what the YouTube API is doing, I suppose this is a reasonable solution.

PR #675, opened 2017-08-18
Show stream details in video information

### LINKED ISSUE #196: Displaying/Grabbing the media URL from Search Results?
I can't seem to workout how to achieve this, but it'd be handy to display or even quick copy the actual URL of videos in search results.

Is this possible or could it be added as feature request?

ritiek:
I splitted part of `_playsong` function in [player.py](https://github.com/ritiek/mps-youtube/blob/8180cedc717b9698a37b5f763183b23c6cfa2f26/mps_youtube/player.py#L138) into `stream_details` function. `stream_details()` is now also called (in [misc.py](https://github.com/ritiek/mps-youtube/blob/8180cedc717b9698a37b5f763183b23c6cfa2f26/mps_youtube/commands/misc.py#L212)) to extract stream information (using `i <number>`) that `mpsyt` is supposed to play.

Closes #196.

### DIFF
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -136,16 +136,7 @@ def _mplayer_help(short=True):
 
 
 def stream_details(song, failcount=0, override=False, softrepeat=False):
-    """ Play song using config.PLAYER called with args config.PLAYERARGS."""
-    # pylint: disable=R0911,R0912
-    if not config.PLAYER.get or not util.has_exefile(config.PLAYER.get):
-        g.message = "Player not configured! Enter %sset player <player_app> "\
-            "%s to set a player" % (c.g, c.w)
-        return
-
-    if config.NOTIFIER.get:
-        subprocess.Popen(shlex.split(config.NOTIFIER.get) + [song.title])
-
+    """Fetch stream details for a song."""
     # don't interrupt preloading:
     while song.ytid in g.preloading:
         screen.writestatus("fetching item..")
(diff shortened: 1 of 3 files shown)


ids1024:
I'm not sure if this information should be displayed in the info for a video, because streams are actually in several formats, although this displays the one it will play. Therefore, `show_video` for instance changes what it displays.

ritiek:
Not sure either but I couldn't find a better place to show this info. Displaying all the stream urls will indeed create a mess. I made a *stream info* section. IMO It should hint people that it will display information about the stream it will play, so they know would know what to expect (mp4, webm or anything else).

Another way could be to add a command like `s <number>` or something that will show stream info. What do you think?

ids1024:
The long stream urls do add clutter for the (probably) majority of uses where that is not the goal of the info command (four lines for me with a full screen terminal; worse with a smaller one).

Another way to provide the functionality of https://github.com/mps-youtube/mps-youtube/issues/196 would be to add a command to copy the url to the clipboard; the `x` command (which requires pyperclip to be installed) copies the video url; perhaps an `X` command that copies the stream url?

ritiek:
Yeah, the `X` command would be a nice way to copy the stream url. I'll put some work to it.

However, what about adding a new `s` command anyway that displays further information (bitrate, extension, etc) about the stream?

ids1024:
> However, what about adding a new s command anyway that displays further information (bitrate, extension, etc) about the stream?

That could be useful as well, so sure, it can be added.

ritiek:
OK, added both `X` and `s` commands. Let me know if anything needs to be changed.

mps_youtube/player.py
```diff
@@ -134,7 +135,7 @@ def _mplayer_help(short=True):
     return lines.format(c.g, c.w)
 
 
-def _playsong(song, failcount=0, override=False, softrepeat=False):
+def stream_details(song, failcount=0, override=False, softrepeat=False):
     """ Play song using config.PLAYER called with args config.PLAYERARGS."""
     # pylint: disable=R0911,R0912
     if not config.PLAYER.get or not util.has_exefile(config.PLAYER.get):
```
ids1024: It should only test for a player when actually playing. Nor should the `NOTIFIER` be called.

ritiek: Yep, missed that.

PR #678, opened 2017-08-25
Load tracks from a Spotify playlist

ritiek:
This PR adds 2 commands; `splaylist` and `suser`.

`splaylist <link to spotify playlist>` loads tracks from Spotify playlist and then attempts to match them on YouTube (similar to album search).

`suser <spotify username>` takes Spotify username and shows all the public playlists owned by the user and lets you choose the playlist you want to load. It then attempts to match tracks on YouTube.

example:

```
splaylist https://open.spotify.com/user/11124160165/playlist/1lKPLKqfYb6k6ENiRfqcVd
```
```
suser sirsirington
```

These features require a new dependency [spotipy](https://github.com/plamere/spotipy), which is wrapper for the Spotify Web API.

I wanted to have this functionality for a long time, thought people might find it useful too.

### DIFF
--- a/mps_youtube/commands/__init__.py
+++ b/mps_youtube/commands/__init__.py
@@ -23,5 +23,5 @@ def command(regex):
 
 
 # Placed at bottom to deal with cyclic imports
-from . import download, search, album_search, misc, config, local_playlist
+from . import download, search, album_search, spotify_playlist, misc, config, local_playlist
 from . import play, songlist, generate_playlist
--- a/mps_youtube/commands/album_search.py
+++ b/mps_youtube/commands/album_search.py
@@ -225,11 +225,6 @@ def search_album(term):
         show_message("Album '%s' not found!" % term)
         return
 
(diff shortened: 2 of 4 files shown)


ids1024:
> These features require a new dependency spotipy, which is wrapper for the Spotify Web API.

Perhaps this should be an optional dependency? That can be achieved by catching `ImportError` and not enabling the functionality if the library is not available.

The `pyperclip` dependency for clipboard support works this way.

ritiek:
Yep, that sounds a good idea. Made spotipy as an optional dependency.

By the way, typing  `suser` without any parameter does not prompt the user to input the parameter (unlike `album`). Similar case with `splaylist`. This probably requires some changes to regex but I am not very familiar with it.

ids1024:
Thanks! https://github.com/mps-youtube/mps-youtube/commit/6b03b67685165790a5fb238d82a16421cf81458a fixes the regex so the commands work with no argument, and prompt the user.

Issue #667, opened 2017-08-11
History extraction


sigismund:
Is there anyway for me to extract (and later download) all history without manually browsing through all pages?

kraetzin:
There isn't an in-app way to do it, but you can unpickle the `play_history` file and wrangle it into a human readable one. Something (hacky) like this would work:
```python
import pickle
history = pickle.load(open('play_history', 'rb'))['history']
with open('human_history.txt', 'w') as f:                                     
    for song in history.songs:
        f.write("%s %s\n" % (song.ytid, song.title))
```
This would write the Youtube ID and song title of every song in your history to `human_history.txt`.

I may have a go at implementing a proper `history export` function for mpsyt if I get time.

sigismund:
Thank you for this snippet. I really appreciate it.

PR #681, opened 2017-09-02
Implement history export function

kraetzin:
This adds `history export` as a way to export the song history to a human readable file called `play_history.txt`. This was requested in #667.

### DIFF
--- a/mps_youtube/commands/misc.py
+++ b/mps_youtube/commands/misc.py
@@ -332,8 +332,9 @@ def clear_history():
 @command(r'history export')
 def export_history():
     """ Export the user's play history to play_history.txt. """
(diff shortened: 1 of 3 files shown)


ids1024:
Perhaps it would be best to just change the history format to a text-based alternative? You just need to be careful to convert the existing one.

The same should be done for playlists and settings (https://github.com/mps-youtube/mps-youtube/issues/112). There was some bike-shedding there over the best formats.

kraetzin:
Yeah I agree that sounds like a better alternative. I'll have a look at implementing the m3u format since that seems to be what was favoured in that issue. I'm closing this pull request and will open a new one for m3u.

Issue #112, opened 2014-09-21
Make portable playlists


icymist:
One problem that I face when making playlists on YouTube is that after spending time to make a playlist, sometimes the videos disappear (video made private, video deleted, etc.) without leaving behind any information about the deleted item.

I think this can be overcome by MPS by doing the following:
1. Add config options to save playlists to a desired location
2. Make playlists portable by writing them in yaml with enough information about the item, so that in case the original video is deleted, one at least has enough information about it to add a new source url. Also, syncing across computer will become feasible.

ids1024:
@np1 Do you have anything against storing playlists and configuration in a json based format? Plain text formats are more user friendly and I personally find no benefit to using pickle.

np1:
No I have nothing against that, I think it would be good :smile:

Atlantic777:
Ok, let's bump this one. Here are my thoughts:

- Items should have ytid, URL or something so we can quickly load them without querying yt API
- In case link is broken (in example, the video is deleted) we should have enough information to search for alternative. A song matcher is already implemented in album search, so we can rely on that one. It uses song duration and title. There's an open ended question how to extract song title from song video.
- A nice feature would be to check playlist integrity (check if all items are available) and search for substitutions (manually or automatically)
- Manual matching if match confidence is low. Let user search for item and pick desired or let him skip that item if alternative is really not available.

I find writing JSON files by hand a bit difficult.  I prefer YAML when there is need for users to build such files manually. In that case, our playlist could be implemented in such way that it can work on playlist with only some information available. In example, only song title is required, and playlist loader should expand and add other fields upon loading the file, so a user can type in the rest or fill it interactively (either manually or automatically).

Also, I implemented a basic playlist reconstruction from text files in #361. I would still like to support that way and I see a few solutions.

1.  those two features implemented separately (I don't like this one)
1. add header, like #mpsyt-raw and #mpsyt-yaml, and trigger needed parser based on that
1. make parser guess which format is used (easier for users but a bit more complex code and I think that it's not worth it)
1. make some sort of convert script (built separate tool, I don't like this either)
1. insist to use only YAML style playlists

ids1024:
@Atlantic777 Yaml is fine, but it would add another dependency. Not that I think that is a major issue, but it is worth mentioning.

You can see what is currently stored in a playlist in https://github.com/mps-youtube/mps-youtube/blob/develop/mps_youtube/playlist.py. For each video, there is the id, title, and length (in seconds).

[...]

PR #682, opened 2017-09-03
Use m3u files for playlists

kraetzin:
The aim here is to switch playlists over from using a single pickled file to using multiple files in the m3u format (see #112).

I have implemented this by creating a folder called `playlists` in the mpsyt directory which expects playlists in the form `<playlist_name>.m3u`.  All existing mpsyt playlists get exported from `playlist_v2` into this directory. The extended `#EXTM3U` format is used whenever mpsyt saves playlists, but it can also handle a simple list of urls. Each m3u file is opened and checked for the header `#EXTM3U`, and if it does not contain this on the first line then a list of urls (or YTIDs) is expected.

This currently needs testing and I'm certain there's many unhandled errors, especially with regards to older mpsyt playlist formats. I'm also planning to switch the `play_history` file to m3u (effectively implements #667 and #681), but would like to make sure this is a sane implementation first.

I don't consider this finished and would love feedback on the changes :smiley:.


ids1024:
This looks good in general. It's hard to check every edge case, but hopefully it can't break anything permanently since the old playlist file is kept.

One small issue: deleting a playlist with `rmp` needs to be handled. With the current implementation, it doesn't remove the file, so the playlist is still there when mpsyt is opened again.

kraetzin:
I believe I've fixed the `rmp` issue. `mv` also works as expected now. I've moved some exceptions from `load()`, that don't apply to the new format, to the conversion function, but I wouldn't know how to test those. I could remove them completely but I figured they do no harm and might help in some edge case.

kraetzin:
I've implemented the history of played songs as m3u and the conversion seems to be working fine. The new file is simply called `play_history.m3u`.

PR #687, opened 2017-09-11
Fix absolute seek through MPRIS with mpv v0.24.0

mg6:
When using an MPRIS-compatible playback widget under GNOME, the
following error appears as the user sets absolute track position:

```
  [ipc_0] Command seek: argument 2 has incompatible type.
```

As it turns out, mpv v0.24.0 expects seek mode argument to be a string.

### DIFF
--- a/mps_youtube/mpris.py
+++ b/mps_youtube/mpris.py
@@ -432,7 +432,7 @@ class Mpris2MediaPlayer(dbus.service.Object):
             Sets the current track position in microseconds.
         """
         if track_id == self.properties[PLAYER_INTERFACE]['read_only']['Metadata']['mpris:trackid']:
-            self._sendcommand(["seek", position / 10**6, 2])
+            self._sendcommand(["seek", position / 10**6, 'absolute' if self.mpv else 2])
 
     @dbus.service.method(PLAYER_INTERFACE, in_signature='s')
     def OpenUri(self, uri):


ids1024:
Will this break the feature on older mpv versions?

mg6:
Dnia 11 września 2017 19:18:59 CEST, Ian Douglas Scott <notifications@github.com> napisał(a):
>Will this break the feature on older mpv versions?

Not sure. I used the latest version available in Ubuntu zesty. Will test this with older releases of mpv and let you know.

-- 
Cheers,
mg6

mg6:
Tested with packaged versions of `mpv` in several Ubuntu releases. Here are the results.
```
0.3.4-1 (trusty):
    seek <sec> absolute     works
    seek <sec> 2            works
0.14.0-1build1 (xenial):
    seek <sec> absolute     works
    seek <sec> 2            fails
0.24.0-1 (zesty):
    seek <sec> absolute     works
    seek <sec> 2            fails
```
Seems like the numeric flag is only supported in very old versions of `mpv`. On the other hand, string seek modes are supported at least since `v0.3.4`. The documentation of `absolute` flag appears as early as in `v0.1.0` ([source](https://github.com/mpv-player/mpv/commit/65074ec1f074d22101b4c8bd95eb4a0354401246#diff-736525780c14c681ed3332cdfa6dcfe0R52)).

I’ll reword the commit title as the bug under discussion is not related to recent versions of `mpv`.

ids1024:
Ah, then it shouldn't be an issue. Thanks!

mg6:
No problem. Thank you, too!

Issue #694, opened 2017-09-24
Can't remove items from playlists


Recchan:
When using the pl function or editing local playlists, I removed some items from the playlist then did a quick next + prev page just to see the item I removed there again.

ritiek:
I can reproduce this problem as well. It seems like a bug with `mpsyt`.

kraetzin:
The bug should be fixed in `develop` now. Hopefully works as expected now. 😄

PR #695, opened 2017-09-25
Fix rm command to correctly remove song from playlist

kraetzin:
Fix for #694. Changes to playlists are now preserved between page changes. This does not affect search results where changes are still reset if you change page, but I'm not sure if that needs changing.


kraetzin:
Hmm this still needs a bit more work. Using `rm` on search results sometimes outputs "Invalid item / range entered!", but does remove the item correctly.

ids1024:
Ah yes, it probably should do that. :laughing: 

Thanks!

PR #699, opened 2017-10-06
commands: search: fix TypeError when search results contain a channel

paulfertser:
When searching for e.g. "Ben Howard" one of the results is a channel and
get_track_id_from_json gets confused:

(Pdb) p get_track_id_from_json(items[9])
{'channelId': 'UC7P46taO0CdI8Gy44P1X2yA', 'kind': 'youtube#channel'}

Make get_track_id_from_json ignore nodes that are obvisouly not track
ids.

Signed-off-by: Paul Fertser <fercerpav@gmail.com>

### DIFF
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -427,7 +427,7 @@ def get_track_id_from_json(item):
         for p in field.split('/'):
             if node and isinstance(node, dict):
                 node = node.get(p)
-        if node:
+        if node and isinstance(node, str):
             return node
     return ''
 


paulfertser:
Duh, was fixed by 4ec28353 apparently (sorry for not testing with HEAD before sending this patch) not sure if that was more correct though, as get_track_id_from_json seems to be written in a way to handle plenty of different cases. Feel free to close this if you think it's meaningless.

Razesdark:
If this issue has been fixed, should we just close it?

paulfertser:
Can some expert please verify that the fix was actually correct and not breaking other parts?

Razesdark:
It shouldn't be someone else job to verify that this PR **isn't** needed; it should be your job to actually prove it actually is. 

Unless you find a case where the #664 breaks, then I would think it is safe to assume that it is correct and does not break other parts.

Even so, I am opposed to having `get_track_id_from_json` just silently ignore erroneous data. We should rather make it so it doesn't receive erroneous data in the first place.

PR #702, opened 2017-10-13
updated youtube-dl notation

sheshang:
here `youtube-dl` was referred as `youtube_dl` buit it is wrong. It will give error. So actually you should write as `youtube-dl`


ids1024:
Hm, I thought `pip3 install youtube-dl` was working (as an some sort of alias, or such). Has something changed? In any case this is probably better.

Thanks!

PR #720, opened 2017-11-10
Remember volume across multiple songs

Razesdark:
I figured we should find a more consistent way of dealing with volume changes across different songs.
Since we use the volume when we draw the player, I've started storing that in a global variable and
use that value to set the player volume between songs.

I've only tested and implemented this for `mpv` and will implement this for `mplayer` as well. 

## Improvements & Testing
What other things should we do to make this a robust feature? Hopefully without adding much complexity for the user. There are some questions that I would like to raise and discuss.

* Does there need to be setting to turn this feature on or off? 
* Should the volume level persist between sessions?
* Should we expose the variable to the playerargs function?
* Are there generic ways to get the volume of non supported players?

If anyone has some ideas regarding how this functionality should work feel free to raise them as well.

### DIFF
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -311,6 +311,8 @@ def _generate_real_playerargs(song, override, stream, isvideo, softrepeat):
                     util.list_update("--really-quiet", args, remove=True)
                     util.list_update(msglevel, args)
 
+            if g.volume:
+                util.list_update("--volume=" + str(g.volume), args)
             if softrepeat:
                 util.list_update("--loop-file", args)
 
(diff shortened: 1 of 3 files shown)


ids1024:
This is probably a good idea.

> Are there generic ways to get the volume of non supported players?

Not really; nor would there be a generic way to set the volume.

Razesdark:
On my setup, mplayer would remember volume by itself.
I've added the code to control it via mpsyt anyways, in case this is some kind of version specific setting etc.

mps_youtube/player.py
```diff
@@ -311,12 +315,15 @@ def _generate_real_playerargs(song, override, stream, isvideo, softrepeat):
...
 
+            if g.volume:
+                util.list_update("--volume=" + str(g.volume), args)
             if softrepeat:
                 util.list_update("--loop-file", args)
 
     elif "vlc" in config.PLAYER.get:
         util.list_update("--play-and-exit", args)
 
+    print("args: " + str(config.PLAYER.get) + str(args) + str(stream['url']))
```
ids1024: There probably shouldn't be a `print` here.

PR #723, opened 2017-11-14
Update ISSUE_TEMPLATE.md

Razesdark:
We seem to be getting predominantly issues where we end up having to ask for the same information; mpsyt version, youtube-dl version etc.

I suggest we add a issue template in an effort to try to raise quality of incoming issues.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,31 +1,11 @@
 <!--- Provide a general summary of the issue in the Title above -->
 
-## Expected Behavior
-<!--- If you're describing a bug, tell us what should happen -->
+## Issue / Suggestion
+<!--- If you're describing a bug, tell us what error you are experiencing -->
 <!--- If you're suggesting a change/improvement, tell us how it should work -->
-
-## Current Behavior
-<!--- If describing a bug, tell us what happens instead of the expected behavior -->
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,31 @@
...
+2.
+
+## Context
+<!--- How has this issue affected you? What are you trying to accomplish? -->
+<!--- Providing context helps us come up with a solution that is most useful in the real world -->
+
+## Your Environment
+<!--- Include as many relevant details about the environment you experienced the bug in -->
+### mpsyt
+<!--- Run "mpsyt --debug" and post your version information here --->
```
ids1024: `mpsyt --version` prints the same information and then exits.

ids1024:
I'm not really that fond of Github issue templates; people often don't use them correctly, just adding noise. And personally I often delete them when submitting issues.

That said, given that Github doesn't have a better alternative, it probably is best to have an issue template at least asking for the output of `mpsyt --version`. But perhaps a more minimal template would be better? I'm not quite sure.

Razesdark:
Something that hopefully gets someone to consider how to actually phrase their issue, asks people to post their stack trace and version information is probably the most important things to accomplish with a template.

PR #737, opened 2017-12-11
Load MP3s from a local directory

ritiek:
This is still not in any state of being merged. It is just a rough implementation. I wanted some general feedback, if this is the right way to do it.

At the moment, it will look only for MP3s in a directory since MP3 track duration can be known using [mutagen](https://github.com/quodlibet/mutagen) but not for other extensions. We could use FFprobe to detect it for all possible media extensions but it won't be a pythonic solution anymore.

To test this out:
```
pip install mutagen
```

```
./mpsyt local open /path/to/mp3/music
./mpsyt local view /path/to/mp3/music
```
Expect some bugs and I've only tested this with mpv-player at the moment.

Also, the regex should be changed to be more restrictive (it works for any argument at the moment) and it may sometimes say that the directory does not exist when using relative paths.

### DIFF
--- a/mps_youtube/commands/local_playlist.py
+++ b/mps_youtube/commands/local_playlist.py
@@ -1,7 +1,9 @@
 import re
+import os
+from mutagen import mp3
 
 from .. import g, c, playlists, content, util
-from ..playlist import Playlist
+from ..playlist import Playlist, LocalMedia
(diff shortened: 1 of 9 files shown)


ritiek:
Closing this as stale.

PR #738, opened 2017-12-12
Fix view command

### LINKED ISSUE #735: What is the difference between `view` and `open` commands?
I took a look at the code in here

https://github.com/mps-youtube/mps-youtube/blob/8cf7c7313b494852b10f0f5cd13fdf5307150a3c/mps_youtube/commands/local_playlist.py#L121-L142

It looks like that it makes the tracks in a playlist *active* when using `open`, but I can do everything with `view` as well, play tracks, save playlist, etc.

Is there supposed to be a difference between them or is this an unexpected behavior?

ritiek:
I don't how `view` command originally worked but now when using `view`, it will temporarily show the expected playlist and then immediately switch back to the last loaded playlist.

Fixes #735.

### DIFF
--- a/mps_youtube/commands/local_playlist.py
+++ b/mps_youtube/commands/local_playlist.py
@@ -133,7 +133,7 @@ def open_save_view(action, name):
             g.active.songs = list(saved.songs)
             g.last_opened = name
             msg = util.F("pl loaded") % name
-            paginatesongs(g.active, msg=msg)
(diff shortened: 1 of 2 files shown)


ids1024:
I think this is probably not the best solution; but I'm not sure what caused this to stop working correctly...

ritiek:
Closing this as discussed in #735.

Issue #495, opened 2016-06-14
[Feature Request] Add 'headers' to a song added to mpd/mpc


faan11:
Hey,
I'm using mps-youtube with a mpd using the way described in the following issue https://github.com/mps-youtube/mps-youtube/issues/168
But i can only see the url of the video as title. I would like to have also the artist name, the title of the song, the album image and so on. This would allow me to distinguish songs. Is there any way?

ids1024:
It might be possible to pass the title of the YouTube video, and the YouTube thumbnail as album art. I do not know how these could be passed to `mpc add` though.

It is not really possible, or at least not easy to get the actual album name, artist, etc. This is addressed in #209.

vn-ki:
This is probably possible with `python-mpd2`.

PR #739, opened 2017-12-13
Show proper metadata including album and artist whenever possible

### LINKED ISSUE #209: artist, album, song title metadata
When I download a song, it would be sweet if the metadata would already be filled out... Is that something thats not too difficult?

vn-ki:
Now fixes metadata when playing songs. This could be extended to fixing metadata (track title, album, artist, album art) when downloading songs. But that would add mutagen as a dependency; so I was not sure of implementing that (That could be a future PR). Switches to old method of displaying thumbnails if no proper metadata was found. Please see attached image. Fixes #209 and #495 .

image

### DIFF
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -635,6 +635,7 @@ def _make_status_line(elapsed_s, prefix, songlength=0, volume=None):
     return prefix + status_line + vol_suffix
 
 def _get_metadata(song_title) :
+    ''' Get metadata from a song title '''
     t = re.sub("[\(\[].*?[\)\]]", "", song_title.lower())
     t = t.split('-')
 
@@ -662,6 +663,7 @@ def _get_metadata(song_title) :
     return metadata
 
 def _get_metadata_from_lastfm(artist, track) :
+    ''' Try to get metadata with a given artist and track '''
     url = 'http://ws.audioscrobbler.com/2.0/?method=track.getInfo&api_key=12dec50313f885d407cf8132697b8712&'
     url += urllib.parse.urlencode({"artist" :  artist}) + '&'
     url += urllib.parse.urlencode({"track" :  track}) + '&'
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -11,6 +11,7 @@ import math
 import time
 import shlex
 from urllib.error import HTTPError, URLError
+import urllib
 
 from . import g, screen, c, streams, history, content, paths, config, util
 
@@ -301,7 +302,7 @@ def _generate_real_playerargs(song, override, stream, isvideo, softrepeat):
             util.list_update("-really-quiet", args, remove=True)
             util.list_update("-noquiet", args)
             util.list_update("-prefer-ipv4", args)
-            
+
 
         elif "mpv" in config.PLAYER.get:
             if "--ytdl" in g.mpv_options:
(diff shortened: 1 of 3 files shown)


ids1024:
> This could be extended to fixing metadata (track title, album, artist, album art) when downloading songs. But that would add mutagen as a dependency

Currently `mps_youtube/commands/download.py` has a very primitive parser (`extract_metadata`) to try to get artist and title from songs, and uses avconv/ffmpeg to add the metadata.

But of course, this solution with lastfm is better, and sharing the metadata through mpris is great too.

mps_youtube/player.py
```diff
@@ -374,14 +375,24 @@ def _get_input_file():
 def _launch_player(song, songdata, cmd):
     """ Launch player application. """
 
+    #song.title
```
hrnr: some debug?

mps_youtube/player.py
```diff
@@ -374,14 +375,24 @@ def _get_input_file():
...
-    arturl = "https://i.ytimg.com/vi/%s/default.jpg" % song.ytid
+    metadata = _get_metadata(song.title)
+
+    if metadata == None :
+        arturl = "https://i.ytimg.com/vi/%s/default.jpg" % song.ytid
+        metadata = (song.ytid, song.title, song.length, arturl, [''], '')
+    else :
+        arturl = metadata['album_art_url']
+        temp = (song.ytid, metadata['track_title'], song.length, arturl, [metadata['artist']], metadata['album'])
+        metadata = temp
```
hrnr: we can assign to metadata directly, right?

hrnr:
Thanks again for this. Seems to work reasonably well for most of the songs. Is usage of last.fm API anyhow limited? I'm just worried if all of our users start to use this, we might get blocked.

For this reason, there should be a way to change the API key. It can be similar to what we have for youtube API key.

vn-ki:
@hrnr I don't think the last.fm API would be problem. Their [terms of service](https://www.last.fm/api/tos) talk nothing about the quantity of requests made. But we will have to give them proper credits for their data. I was thinking about implementing this for downloads too. Is dependency on mutagen a problem?

mps_youtube/player.py
vn-ki: Ya! Sorry.

mps_youtube/player.py
vn-ki: I don't know why I did that. XD

hrnr:
Thanks. I have tested the MPRIS side and all looks good now (all datatypes correct etc.).

ids1024:
> Is dependency on mutagen a problem?

I'm hesitant to add non-optional dependencies, but mutagen seems to be commonly packaged by distributions (Arch and Debian, at least), so it should be fine.

Razesdark:
We could also implement the musicbrainz project as a source for music information instead of last.fm. 
Musicbrainz is an open source project and the data is released under CC0, without any need to mention the musicbrainz project when using the data. 

It also does not require an API key so the functionality won't break if the hardcoded key is revoked in the future. We're also not attributing them when using their data(we also need to show a link to https://last.fm).

vn-ki:
@Razesdark Do they provide the album art with the json response? Their web service don't expose many things. Is there a way to search using both title and artist? I tried a few songs and the top result don't match (Which means more work to find the correct result). If there is way to search both title and artist this would improve. But still I couldn't see the ablum art in the json response. I have seen many open source projects using last.fm (Jockey and Phonograph, to name some).

Issue #135, opened 2014-11-01
iTunes Not Playing Files


josh-byster:
Hi,

I first want to say thank you for providing such a great open source package. I'm not sure if this is just a "me" issue or a general one. I've downloaded a few different audio (.m4a) files and tried adding them to iTunes through various methods. The time was 0:00 in iTunes, and it said "Not available". Is anyone else having this issue?

Thank you!

PR #740, opened 2017-12-14
Added fixing of metadata when downloading m4a audio files.

vn-ki:
This PR is in continuation with #739 . This PR adds mutagen as an optional dependency. If mutagen is found and the file currently downloading is a m4a audio file, then it tries to fix the metadata.

### DIFF
--- a/mps_youtube/commands/download.py
+++ b/mps_youtube/commands/download.py
@@ -382,7 +382,7 @@ def _download(song, filename, url=None, audio=False, allow_transcode=True):
     ext = filename.split(".")[-1]
     valid_ext = ext in active_encoder['valid'].split(",")
 
-    if audio and filename.split('.')[-1] == 'm4a':
+    if audio and filename.split('.')[-1] == 'm4a' and MUTAGEN_PRESENT:
         remux_audio(filename, song.title)
 
     if config.ENCODER.get != 0 and valid_ext and allow_transcode:
--- a/mps_youtube/commands/download.py
+++ b/mps_youtube/commands/download.py
@@ -6,7 +6,14 @@ import shlex
 import random
 import subprocess
 from urllib.request import urlopen
+import urllib
 from urllib.error import HTTPError
+
+try :
+    from mutagen import mp4
+    MUTAGEN_PRESENT = True
+except :
+    MUTAGEN_PRESENT = False
 
 from .. import g, c, screen, streams, content, config, util
 from . import command, PL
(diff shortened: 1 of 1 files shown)


vn-ki:
@ids1024 What does the remux_audio function do? I'm not sure what remuxing is and whether it's needed or not.

ids1024:
It seems the remuxing code was added in 32e9739f to fix #135. The metadata handling was added to that function later in a96393c1a5ab49a5e7019722c1b976407e305f0f.

vn-ki:
@hrnr Can you, or anyone, please review this PR?

vn-ki:
Can someone review this PR, @ids1024 , @hrnr . Sorry for the bugging.

mps_youtube/commands/download.py
```diff
@@ -231,38 +238,14 @@ def _make_fname(song, ext=None, av=None, subdir=None):
...
-    meta = extract_metadata(title)
-    metadata = ["title=%s" % meta["title"]]
 
-    if meta["artist"]:
-        metadata = ["title=%s" % meta["title"], "-metadata",
-                    "artist=%s" % meta["artist"]]
+    cmd = [g.muxapp, "-y", "-i", temp_file, "-acodec", "copy", "-vn", filename]
 
-    cmd = [g.muxapp, "-y", "-i", temp_file, "-acodec", "copy", "-metadata"]
-    cmd += metadata + ["-vn", filename]
```
hrnr: Please do not destroy writing metadata with ffmpeg.

hrnr:
Hi I took a look. Rater than introducing Mutagen, I suggest different approach. `ffmpeg` can already write metadata to mp4 files see https://wiki.multimedia.cx/index.php/FFmpeg_Metadata. We are actually using that right now.

You can extend `remux_audio` to fetch proper metadata with `util._get_metadata` and write it with `ffmpeg`. Just add more command line options to ffmpeg for needed metadata (as it is already done now). See the link above. If metadata can't be downloaded, I would suggest setting title to youtube title to be consistent with what we do in MPRIS.

If you also want to add album cover, that can also be with `ffmpeg`, but this might be more tricky see https://stackoverflow.com/questions/18710992/how-to-add-album-art-with-ffmpeg.

This way we can avoid mutagen dependency and do everything in just one step.

What do you think? Does this approach makes sense to you?

vn-ki:
Hi @hrnr . Thank you for your review. I did try to add metadata to the audio file using ffmpeg. I did not want to add a dependency just to add metadata to the downloaded file. But ffmpeg seems to only support adding album art to mp3 files. Trying to add album art to a m4a file returns an error. I have no experience with ffmpeg so I don't know how to work around this. I couldn't find any official sources saying ffmpeg can add album art to m4a files. I'm adding a screen shot of the error if you could help me with this.
image

hrnr:
Might be possible that ffmpeg does not support that for m4a container. There is an open issue https://trac.ffmpeg.org/ticket/2798.

Then I think we should write as much as possible with ffmpeg (all users will benefit). For album arts then we can use mutagen if available.

You can also decide to leave album arts for now and wait a few years for fixed issue in ffmpeg :)

vn-ki:
Haha. Okay. That is a 4 year old issue. But what you said is true; using ffmpeg would benefit most users. But mutagen is a more pythonic way(than using subprocess). That's why I went with mutagen initially. 

Maybe I should close this PR and open a new one.

Thank you.

PR #748, opened 2017-12-28
Player_status: minutes are now correctly displayed

Laxa:
Code is kind of incoherent, minutes can go above 60 and then they are correctly displayed when above 100.


ids1024:
I don't know why it was like that, but this should be correct. Thanks!

Issue #142, opened 2014-11-03
Last.fm scrobbling support


andrewmy:
It would be a very nice feature to scrobble the tracks.
Posting to the API is quite easy, not so sure about authentication.

ids1024:
Authentication with last.fm requires an api key.  Should the user provide their own?  An api key included with mpsyt would be visible to anyone.

http://www.last.fm/api/scrobbling
http://www.last.fm/api/authentication

np1:
Another consideration is that a lot of items played through mps-youtube are not songs or items that should be scrobbled - how do we determine which items should be scrobbled? How do we determine the artist and track names with confidence? Perhaps if this is implemented it should only apply to items that were retrieved using the `album` command.

w-hive:
+1 for this!

You could probably add the support in easily with pyscrobbler

https://github.com/offmessage/pyscrobbler

PR #758, opened 2018-01-12
Added Last.fm support

rien333:
Implemented optional Last.fm scrobbling support (closing #142) using the [pylast](https://github.com/pylast/pylast) module. Although `util.py` contains a piece of code that uses the Last.FM web API, I opted for the pylast module because it is well maintained and easily extendible to Libre.fm. I figure the aforementioned web API call can be replaced by a pylast call. 

I tried to to pay attention to the concerns raised in #142. Namely, first of all, scrobbling is only enabled for the `album` command. This seemed like a good idea because `album` always provides accurate song metadata and furthermore ensures that the user wants to listen to music (although the 'search music' config option is reflective of this as well). 

Because scrobbling is only supported for the `album` command, I had to monitor the user input for it's usage within `main.py`. Not the cleanest solution, but I couldn't think of an alternative given the current code structure. I also had to hijack the player and a few other things, so the footprint of this commit is not super small. It might be possible to read the (readline?) history instead of monitoring input in `main.py`, but I am not completely convinced of this idea.

The second concern raised in #142 is how the user will provide Last.fm credentials and API information. This is implemented with the `set` command which can store this information om the standard configuration file. You can inspect some details of the  workings of the Last.fm variables and the authentication via the `help` command. Because last.fm support is optional, these variabeles remain unused until the user provides all required information. You can then connect to last.fm either automatically at startup or by running a new command called `lastfm_connect`, which also provides potential pylast error messages.

It would probably have been more fitting to have implemented this as a plugin, but the plugins feature of this projects seems to be pretty experimental and possibly abandoned.

A final worry is that the `commands/lastfm.py` file contains some helper functions that would be more  appropriately fitted in `util.py`. Any input is appreciated (I am also new to opensource projects of this scope).


vn-ki:
Wouldn't it be nicer if pylast is added as an optional dependency?

rien333:
@vn-ki Yes! That is how I tried to program it. I tested this PR without pylast installed, and it seems that it still works fine. What do you mean exactly?

ids1024:
>It would probably have been more fitting to have implemented this as a plugin, but the plugins feature of this projects seems to be pretty experimental and possibly abandoned.

Agreed; it would be better as a plugin, but plugin support was never finished, so that's not an option.

mps_youtube/config.py
```diff
@@ -302,6 +314,10 @@ class _Config:
             ConfigItem("window_size", "",
                 check_fn=check_win_size, require_known_player=True),
             ConfigItem("download_command", ''),
+            ConfigItem("lastfm_username", ''),
+            ConfigItem("lastfm_password", '', check_fn=check_lastfm_password),
+            ConfigItem("lastfm_api_key", ''),
+            ConfigItem("lastfm_api_secret", ''),
```
ids1024: Adding not only one but four configuration options just for this isn't great, although I don't know if there's a better way to do it. I guess it might be better if mpsyt had multiple sections for settings or something like that...

This could by an issue if the list output by `set` doesn't fit on a page; it isn't currently paginated. Of course, that issue isn't specific to this change.

rien333: Oh yes that reminds me, I found 4 options to be a lot as well. Unfortunately, I don't seen any distinctly better solutions either. Maybe the information could be read from a config file ? I think [cmusfm](https://github.com/Arkq/cmusfm) does it that way. It could for example read the last.fm information from a file in `$HOME/.config/mps-youtube/`. 

There is also some ENV variables pylast looks for/sometimes uses in example code, so maybe we could use those (you would have to do something like `export PYLAST_USERNAME=foo`, for example). On the other hand, seemingly nobody else seems to use those variables.

rien333:
Is there any possibility of this/pylast being integrated? If so, I could try to start a PR that ports the  `_get_metadata_from_lastfm` function from the `util.py` file over to pylast. It now uses the Last.Fm web API instead, which is probably subject to change given the "new" Last.Fm beta. I also had a weird issue with it on my raspberry pi, which was related to the json and urllib modules not working properly. The usage of these two modules might be avoided by using pylast instead.

ids1024:
I guess this can just be merged. Although I don't like the four separate configuration options, the alternatives seem somewhat worse. The changes to `main.py` and `play.py` are not ideal, but are fairly minimal and probably the best solution for now. Thanks, and sorry for not merging earlier!

Yeah, it probably would be a good idea to change `_get_metadata_from_lastfm` to use pylast.

vn-ki:
> Yeah, it probably would be a good idea to change _get_metadata_from_lastfm to use pylast.

I think that is a bad idea. It would add dependency to just have the metadata on mpris. What if a user wants metadata but don't want other last.fm features?

PR #769, opened 2018-01-31
Added note about Fedora copr repository

moosingin3space:
I made a Copr for Fedora packages for mps-youtube: https://copr.fedorainfracloud.org/coprs/moosingin3space/mps-youtube/. This pull request adds a description to the readme explaining how to use it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,6 +112,17 @@ You can install mps-youtube directly from the official repositories:
 
     [sudo] apt install mps-youtube
 
+Fedora
+~~~~~~
+You can install mps-youtube from a `copr <https://copr.fedoraproject.org>`_::
+
+    [sudo] dnf copr enable moosingin3space/mps-youtube
+    [sudo] dnf install python3-mps-youtube
+
+This package comes with mpris2 support via its dependences. It also depends on
+youtube-dl and mpv. Note that the mpv dependency requires that you have `RPM
+Fusion <https://rpmfusion.org>`_ free enabled.
+
 macOS X
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Install mpv (recommended player) with `Homebrew <http://brew.sh>`_::


ids1024:
I'm not really familiar with Fedora, but is it really normal to call a package like this `python3-mps-youtube`? I don't know about Fedora conventions, but on Arch Linux it is normal to prefix `python-` to a package's name for Python libraries, but not for applications.

ssieb:
It's the same in Fedora. If it's not a library, there's no need to have python in the package name.

SanketDG:
mpv shouldnt be a dependency here

mpv is an external tool used by `mpsyt`, it should be upon the user to configure it properly

moosingin3space:
Re: package name, I had generated the initial RPM spec using `pyp2rpm`, which automatically prefixes packages. I can adjust it easily.

ids1024:
I do think the package should be renamed to just `mps-youtube` then. Could `mpv` be made an optional dependency (I presume RPM packages have something like that)? it is recommended, but `mplayer` is also supported, and other unsupported players work to varying degrees.

moosingin3space:
RPM has two types of optional dependency, `Recommends`, which by default, `dnf` will try to resolve, and `Suggests`, which `dnf` will not resolve by default. Which do you think is appropriate here?

ids1024:
Hm...

> Note that the mpv dependency requires that you have `RPM Fusion <https://rpmfusion.org>`_ free enabled.

So mpv is not in the official Fedora repositories? In that case, I suppose `Suggests` is better. On Arch, the mps-youtube package (in the official repositories) has an optional dependency on both mpv and mplayer, and [prints a message](https://git.archlinux.org/svntogit/community.git/tree/trunk/mps-youtube.install?h=packages/mps-youtube) about requiring one of these on install.

iamtalhaasghar:
As mps-youtube have been rebranded as `yewtube`. I request you to create a new pull request if you want to maintain your fedora copr.

PR #785, opened 2018-02-21
Fix json error in some python distributions

vn-ki:
Mentioned in #752. ([This comment](https://github.com/mps-youtube/mps-youtube/issues/752#issuecomment-367111112))

### DIFF
--- a/mps_youtube/util.py
+++ b/mps_youtube/util.py
@@ -527,7 +527,9 @@ def _get_metadata_from_lastfm(artist, track):
         metadata['artist'] = data['track']['artist']['name']
         metadata['album'] = data['track']['album']['title']
         metadata['album_art_url'] = data['track']['album']['image'][-1]['#text']
-    except:
+    except KeyError:
+        return None
+    except IndexError:
         return None
 
     return metadata
--- a/mps_youtube/util.py
+++ b/mps_youtube/util.py
@@ -518,7 +518,7 @@ def _get_metadata_from_lastfm(artist, track):
 
     metadata = dict()
 
-    data = json.loads(resp.read())
+    data = json.loads(resp.read().decode('utf-8'))
 
     if 'track' != list(data.keys())[0]:
         return None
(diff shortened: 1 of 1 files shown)


jahlborn:
Another helpful change would be to move the "try:" in _get_metadata_from_lastfm() up to encapsulate the whole method. that way future issues getting metadata won't cause the player to fail.

mps_youtube/util.py
```diff
@@ -518,7 +518,7 @@ def _get_metadata_from_lastfm(artist, track):
 
     metadata = dict()
 
-    data = json.loads(resp.read())
+    data = json.loads(resp.read().decode('utf-8'))
```
ids1024: It seems like the `HTTPResponse` returned by `urlopen` can be passed to `json.load`. That would probably be better...

vn-ki: I have made the change.

jahlborn:
fyi, the alternate version of the patch does _not_ fix the original problem (using load(resp) instead of the loads with decode).  i tested the latest patch the it fails.

vn-ki:
@jahlborn Thanks for the update.

@ids1024 Should we switch back to `json.loads` or is there a workaround with `json.load`?

ids1024:
Does anyone know why this is happening? The following code works for me with Python 3.6. Is it an issue with older Python 3 versions?

```python
import urllib.request
import json
print(json.load(urllib.request.urlopen("https://api.github.com/")))
```

For that matter, `json.loads(b"{}")` works fine; it accepts a bytes object.

ids1024:
Okay, it seems it is a Python version issue. 

From the [docs for `json.loads()`](https://docs.python.org/3.6/library/json.html):

> Changed in version 3.6: s can now be of type bytes or bytearray. The input encoding should be UTF-8, UTF-16 or UTF-32.

Changing it back to `json.loads` is probably the only way to be comparable (or at least, no other way is simpler or better).

PR #797, opened 2018-02-26
fixed #455

### LINKED ISSUE #455: overriding settings from the command line should only affect the current invocation
The `mpsyt set... , set...` feature is very helpful, but it seems to change mpsyt settings permanently, and this can lead to unexpected results in subsequent invocations of the program.

Scenario: as I do for many UNIX programs, I wrote a shell script wrapper for mpsyt, for a very specific need which requires overriding the standard settings via command line parameters. Currently, if I run mpsyt directly, after running my wrapper, mpsyt will keep using the special settings, instead of the expected standard settings.

The expected behaviour, like for standard UNIX programs, is that overriding settings by command-line options only affects the current instance of the program, without permanently changing settings in the program configuration file.

BlackCapCoder:


ids1024:
This will have the effect that any time any command is passed as an argument to mpsyt (not just `set`), changes to settings will not be persisted. Not only changes made by `set` commands that appear in the command line, but also those issued afterwards. This seems undesirable.

BlackCapCoder:
@ids1024 This is intended

BlackCapCoder:
@ids1024 To clarify, this makes it so that no command passed via the program arguments will persistently change settings, be it `set` or not. If the user then manually changes a setting, all settings including the ones set automatically will be saved.
Alternatively one could remember what settings were changed automatically and only save the ones that wasn't. If this is better behavior I will gladly implement it.

ids1024:
> If the user then manually changes a setting, all settings including the ones set automatically will be saved.

Ah, yes. I was having trouble testing this due to network issues, and I was thinking looking at the code that this wasn't the case, but since it's in a loop, `config._automatic` does get set to `False` later.

> Alternatively one could remember what settings were changed automatically and only save the ones that wasn't. If this is better behavior I will gladly implement it.

I've thought of the same thing. Another way to do this would be to provide another way to temporarily set options, which might be better....

BlackCapCoder:
I removed `config._automatic` and implemented an `auto_save_settings` setting instead. When `False` mpsyt will not save setting changes automatically. `auto_save_settings` defaults to `True` as to not break backwards compatibility. Changes to the `auto_save_settings` setting will be always be automatically saved, whether it is enabled or disabled.

I also added a `savesettings` command to allow saving the settings manually.

kraetzin:
With #803 merged, is this functionality still desired? `set` and `set -t` seems to be a more flexible approach. Perhaps we should close this PR?

BlackCapCoder:
Yes, this was just a second alternative

PR #802, opened 2018-03-06
README : Invite to upgrade youtube-dl.

Rom1deTroyes:
In disto like Debian stable, pip3 install the same version as apt does (Feb 2017)...
Upgrading youtube-dl by it-self seems to be mandatory, then.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -105,6 +105,10 @@ Installing youtube-dl is highly recommended::
 
     $ pip3 install --user youtube-dl
 
+Upgrading youtube-dl could help (especially on distros such as Debian where the package in the repos is old)::
+
+    $ youtube-dl -U
+
 For mpris2 support, install the python bindings for dbus and gobject::
 
     $ pip3 install --user dbus-python pygobject


Vrihub:
FYI, I can't reproduce this. I regularly upgrade youtube-dl via pip3 on Debian and I always get the latest release.

ids1024:
Are you passing `-U` to pip? Without that, it won't upgrade.

amartin83:
I guess it's good to point out, that there seem to be 3 different versions of "youtube-dl":

1) from distro (ubuntu, debian, mint) which is usually old
2) from PIP
3) from main website. Instruction says to put it in: /usr/local/bin/youtube-dl

And sometimes there is a difference in its name like "youtube-dl" vs "youtube_dl".

As I'm using youtube-dl for some other apps, so best way would be, to have just one youtube-dl in the system - for me prefered one is from the main website as it's quickest to be updated and also easiest to update.

Question: is it possible (or will this option be possible in future version of mpsyt) to specify which youtube-dl mpsyt should be using? Mayby to specify a path or it automatically look in default folders of the above 3 points?

P.S.
Thanks for a great app, using it all the time and love how well the application has been thought-out

Kind regards,
AMartin

pirate486743186:
wait a minute. If i remember correctly mps can't use the youtube-dl downloaded directly from their site. mps uses youtube-dl as a python module, and for that, it must be installed with pip or from the distribution.

I think you are just using the distribution version or falling back on pafy, i had the joke earlier... If you remove the distribution version it should stop working...

Also "-U" doesn't work with the pip and distribution version.

So.... only  the pip version is actually viable...
The readme should mention some of this.
and also mention how to upgrade it
"pip3 install  youtube-dl --upgrade"

PR #803, opened 2018-03-06
Implemented temporary settings

BlackCapCoder:
This adds the concept of temporary settings. Temporary settings will override the regular settings, but they only apply to the current instance- ie they are not saved to disk.
Temporary settings can be set with the `set -t` command. If the regular `set` command is used, any temporary setting for that key will be discarded.

The current active temporary settings are displayed at the `showconfig` screen, which looks like this:

The temporary column is only visible if there are active temporary settings.

### DIFF
--- a/mps_youtube/commands/config.py
+++ b/mps_youtube/commands/config.py
@@ -6,9 +6,24 @@ from . import command
 def showconfig():
     """ Dump config data. """
     width = util.getxy().width
-    width -= 30
-    s = "  %s%-18s%s : %s\n"
-    out = "  %s%-18s   %s%s%s\n" % (c.ul, "Key", "Value", " " * width, c.w)
+    longest_key = 17
+    longest_val = 0
+    has_temps = False
+
+    for setting in config:
+        val = config[setting]
+        longest_val = max(longest_val, len(str(val.display)))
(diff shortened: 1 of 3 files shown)


Razesdark:
These changes should also be reflected in the help text

BlackCapCoder:
@Razesdark I created a new pull request for this https://github.com/mps-youtube/mps-youtube/pull/848

mps_youtube/commands/config.py
```diff
@@ -6,9 +6,24 @@
...
+        val = config[setting]
+        longest_val = max(longest_val, len(str(val.display)))
+        has_temps = has_temps or val.temp_value is not None
+
+    width -= 27
+    s = f"  %s%-{longest_key}s%s : %-{longest_val+1}s"
+
+    if has_temps:
+        width -= longest_val + 5
+        out = (f"  %s%-{longest_key}s   %-{longest_val}s %s%s%s\n") % (
```
vn-ki: @BlackCapCoder You seemed to introduce python 3.6+ f-strings to the repo. Even though I love this feature, not everyone is on python 3.6 as of now.

Can you send in a quick PR to change this and use `.format` instead?

If you're busy, then that's fine.

PR #805, opened 2018-03-07
Fetch private playlists when using direct spotify link

ritiek:
This PR uses a more direct way and is now able to fetch private playlists when using Spotify URL.

For example, a private playlist:
```
> splaylist https://open.spotify.com/user/spotify/playlist/37i9dQZEVXcT9cQElPZqqw
```
This would fetch tracks from some other (incorrect) playlist, this PR fixes this issue.


ids1024:
This should be better. Thanks!

Issue #799, opened 2018-03-01
No playback, mpsyt returns to the list of found records.


ZaxonXP:
Mpsyt goes back to the list of found videos without playing anything.

## Issue
When I try to play any of the video/audio then mpsyt tries to get the file, but the moment it should run the player it goes back to the result list. I tried mpv/mplayer/vlc players and none of them works.

Please advise how to debug the issue.

## My Environment
OS: Debian 8.10 Jessie
mpsyt: v0.2.8
mpv: 0.8.3
vlc: 2.2.8

vn-ki:
Mpsyt is probably running into an error while fetching the steam and trying to play the item. The log would be the obvious way of debugging. 

Does this happen with every video? Try scrolling up to see if there are any error messages you missed.

ZaxonXP:
How can I get the log file?

I scrolled up, but there was no any error.

Paz666:
Same problem here, It doesn't happen with every video, but it does with a lot of them.

vn-ki:
Open mpsyt with `mpsyt --debug`. Log can be found in the file `/tmp/mpsyt.log`.

ZaxonXP:
Here is what the log produced:

vn-ki:
This is most probably related to #753. This is already fixed in the `develop` branch. Try that version to see if your problem is fixed.

@ids1024 You should probably push that commit to the master branch and release a new version as this bug is affecting many users.

ZaxonXP:
OK, tried to install the dev version with the following command:

`pip3 install --user -U git+https://github.com/mps-youtube/mps-youtube.git`

The installation was without an error, but execution complains about player import: 
[mps_err.log](https://github.com/mps-youtube/mps-youtube/files/1781464/mps_err.log)

vn-ki:
That is bug! It's due to a cyclic import in `misc.py` caused by `lastfm.py`. I will fix it and open a PR soon.

dgalaktionov:
I've been using some old version just fine, but today I've installed 2.8 from pip in another computer and this happened to me. Trying to install it from the repository like @ZaxonXP did gave me the same error on execution.

So I've worked around this issue and others [in my own fork](https://github.com/dgalaktionov/mps-youtube/commit/8cbb3fd7e30c9ad964ea847219abf4522e83a0d6). I am not going to make a PR because instead of sorting out the `lastfm` dependency problem I've just removed it along with its scrobble feature (whatever it means).

[...]

PR #807, opened 2018-03-08
Fix cyclic import and #806

### LINKED ISSUE #806: q doesn't stop for repeat songs
Play any song on repeat, and press `q`, it will just restart the song.

## Your Environment

mpsyt `develop` branch

Fedora 27

mpv

vn-ki:
Fix the import problem mentioned in [this](https://github.com/mps-youtube/mps-youtube/issues/799#issuecomment-370457169) comment in #799 and fix #806.

### DIFF
--- a/mps_youtube/commands/lastfm.py
+++ b/mps_youtube/commands/lastfm.py
@@ -1,6 +1,3 @@
-import time
-import datetime
-
 try:
     import pylast
     has_pylast = True
@@ -9,6 +6,7 @@ except ImportError:
 
 from .. import g, util, config
 from . import command
+
 
 @command(r'lastfm_connect')
 def init_network(verbose=True):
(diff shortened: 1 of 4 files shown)


vn-ki:
@ids1024 Probably not the best place to ask this, but in this code snippet, why is there a `screen.reset_terminal()`? In my machine this line messes up the whole color palette of the terminal (`tset -c` does it too). The code works (at least, on my machine) perfectly without that line. 

https://github.com/mps-youtube/mps-youtube/blob/18fa860f7517099634c453a5b4be3cf15b268e93/mps_youtube/player.py#L77-L83

ids1024:
> Probably not the best place to ask this, but in this code snippet, why is there a `screen.reset_terminal()`?

I think the concern is that the mode of the terminal could be messed up if playing is just terminated; the code is probably there because of an actual issue with that happening.

Perhaps there's a better solution, but I'm not too familiar with `stty`, etc.

ids1024:
I don't really like moving lastfm code out of `lastfm.py`. Another solution to cyclic imports like this is to move the `import` into the function that needs it, though that's ugly here since several functions in `player.py` use it.

Ideally, those calls in `player.py` would be replaced with some generic hook, that plugins like a lastfm plugin could use.

vn-ki:
I think `player` shouldn't import any`command`. That's why I did this. What do you propose we do?

ids1024:
For now, perhaps the `if g.scrobble:` code that is repeated could be replaced with a method call, `BasePlayer.start_playing_hook`, or something like that. Then call `import lastfm` in that function, to deal with the cyclic import.

In the future, this would in theory be replaced with the plugin API, so plugins can register hooks to be called there...

PR #835, opened 2018-05-31
Use maximum available resolution for thumbnail

ritiek:
This PR improves thumbnail quality in metadata when using mpris API (instead of always falling back to `default.jpg`). One drawback is that this makes things a little slower (checking if URLs responds with 200 or not). Not sure if it is worth it though, let's decide!

Cross reference: mps-youtube/pafy#211

### DIFF
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -138,11 +138,28 @@ class BasePlayer:
         """ Launch player application. """
         pass
 
+    def _getbestthumb(self):
+        part_url = "http://i.ytimg.com/vi/%s/" % self.song.ytid
+        # Thumbnail resolution sorted in descending order
+        thumbs = ("maxresdefault.jpg",
+                  "sddefault.jpg",
+                  "hqdefault.jpg",
+                  "mqdefault.jpg",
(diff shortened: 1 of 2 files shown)


ids1024:
Perhaps change this to call the method from pafy? That means it pulls in pafy's `develop` branch as a dependency (and we'll need to release a new pafy before/when releasing a new mps-youtube), but I think this is generally a reasonable thing to do.

ritiek:
Ok, I've changed it to use `Pafy.getbestthumb()`. Since most people have `youtube-dl` installed when using mps-youtube, so there isn't going be any additional delay to fetch the best thumbnail https://github.com/mps-youtube/pafy/pull/211#discussion_r192663977.

Also, I think we should release a new version of pafy before merging this otherwise we'll get half the issues complaining about `Pafy.getbestthumb` method not found :D

vn-ki:
Let's just remove the pafy internal backend. I don't think anyone uses it anymore.

ritiek:
Actually, it still lags (about 3s) probably because of `util.get_pafy`. I'll have to see more into this.

Issue #828, opened 2018-05-19
[Feature Request] List playlists from a channel


thezealousfool:
I would like to see all the playlists from a given channel.

I can search for a channel, using ```channels <channel_name>``` I can then enter into the required channel and all the videos are split out. How about I want to see the playlists in that channel?

maln0ir:
Any progress on this issue? I would be great to have command like `channelpl` or similar, because sometimes finding username is not so obvious.

yardenac:
How come this was never merged? I don't know what the best approach is either, but unintuitive is better than impossible!

PR #840, opened 2018-06-04
Adds ability to search for playlists from channels based on Channel ID

Razesdark:
As it stands our user playlist functionality is kind of obtuse.
 
`userpl` will only let you search by an users username and not the displayed name of the channel, which most certainly confuses a lot of people.  None of the api segments we currently request contain any unique identifier for a user except the channel id, which is a 24 char alphanumeric string and the displayed name which is not unique enough to be searched on to produce a single result, and the unique identifier we currently can search on is not present in any of our search results. Further more, it seems youtube is moving away from unique user identification and subscriptions will now redirect to links that use the unique channel identifier instead of the user it is associated with. 

My current solution takes the unique channel identifier and exposes that in our `channels` search which then can be inputted into a `chanpl <channel_id>` to recieve playlists for a specific channel.

It is not a pretty solution, but it does give us a way to search based on channels and not only users, and a in client way to get all the information needed.

Ref: #828

### DIFF
--- a/mps_youtube/commands/config.py
+++ b/mps_youtube/commands/config.py
@@ -16,14 +16,14 @@ def showconfig():
         has_temps = has_temps or val.temp_value is not None
 
     width -= 27
-    s = f"  %s%-{longest_key}s%s : %-{longest_val+1}s"
+    s = "  %s%-{0}s%s : %-{1}s".format(longest_key, longest_val+1)
 
     if has_temps:
         width -= longest_val + 5
-        out = (f"  %s%-{longest_key}s   %-{longest_val}s %s%s%s\n") % (
+        out = "  %s%-{}s   %-{}s %s%s%s\n".format(longest_key, longest_val) % (
             c.ul, "Key", "Value", "Temporary", " " * width, c.w)
     else:
-        out = f"  %s%-{longest_key}s   %s%s%s\n" % (c.ul, "Key", "Value", " " * width, c.w)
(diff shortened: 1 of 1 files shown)


kraetzin:
I agree with @vn-ki. If possible it would be nice for a user to be able to use `user` and `userpl` for usernames and channel id's. If the channel ID is always 24 characters then it should be reasonable to identify what the user wants?

Razesdark:
Oh yeah, its super easy to identify a channel id and act accordingly.

I'm against using `userpl` because I think it muddies its function by having it work different with one kind of input compared to another. Who would find it natural to search by  *channel_id* using the command `user`. I think the natural expectation of user and userpl is to supply a username and get the results based on that. And we should keep it like that. 

In an example where a user would get all its information from within mpsyt they would have to run the `channels` command to get the channel id, then `user` or `userpl` to use it, that sounds kind of unintuitive to me.

I can see the argument that a layman would not make the distinction between a user and a channel. I would assume that most people think of a channel's name when they see the current `user` command and gets confused why its not working as intended. I wouldn't be opposed to catch both channel names and usernames by the user function as long as the command would react similar in both cases.

kraetzin:
Indeed, I do think the distinction is confusing to laymen. I've tested a bit and it seems like the `channels` command will successfully search for users as well as channel names, so perhaps it's actually `user` and `userpl` which are redundant here. Maybe those could be depreciated in favour of `channels` and `chanpl`? I suppose that would be a seperate PR and discussion to be had :grin:. I agree that yours is a good solution until then.

PR #844, opened 2018-06-10
fix UnicodeEncodeError on latin-1

JervisTetch:
https://github.com/mps-youtube/mps-youtube/issues/843

### DIFF
--- a/mps_youtube/history.py
+++ b/mps_youtube/history.py
@@ -36,7 +36,7 @@ def save():
         hf.write('#EXTM3U\n\n')
         if 'history' in g.userhist:
             for song in g.userhist['history'].songs:
-                hf.write('#EXTINF:%d,%s\n' % (song.length, song.title))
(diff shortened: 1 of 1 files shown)


mps_youtube/history.py
```diff
@@ -36,7 +36,7 @@ def save():
         hf.write('#EXTM3U\n\n')
         if 'history' in g.userhist:
             for song in g.userhist['history'].songs:
-                hf.write('#EXTINF:%d,%s\n' % (song.length, song.title))
+                hf.write('#EXTINF:%d,%s\n' % (song.length, song.title.encode('utf-8')))
```
ids1024: This doesn't look right to me.

Look at the behavior of this code, for instance: `"encoded: %s" % ("test".encode("utf8"),)`. This results in the string `"encoded: b'test'"`. So it produces a str (not bytes) containing `b'` and `'`.

I'm not sure the best solution here. If we want to save the file as UTF-8, then we can pass an `encoding=` argument to `open`.

PR #848, opened 2018-06-13
Added `set -t` to the help text

BlackCapCoder:
Another option is something like:
```
set -t - Temporarily change a setting
```
but I think this looks a little confusing

### DIFF
--- a/mps_youtube/helptext.py
+++ b/mps_youtube/helptext.py
@@ -239,8 +239,8 @@ def helptext():
     {2}set api_key <key>{1} - use a different API key for accessing the YouTube Data API
     {2}set set_title true|false{1} - change window title
     
-    {2}Additionally, set -t may be used to temporarily change a setting without
-    {2}saving it to disk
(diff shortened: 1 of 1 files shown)


mps_youtube/helptext.py
```diff
@@ -238,6 +238,9 @@ def helptext():
     {2}set video_format <auto|mp4|webm|3gp>{1} - set default music video format
     {2}set api_key <key>{1} - use a different API key for accessing the YouTube Data API
     {2}set set_title true|false{1} - change window title
+    
+    {2}Additionally, set -t may be used to temporarily change a setting without
+    {2}saving it to disk
```
vn-ki: Any specific reason for the whitespace? Why make it two line instead of one?

Also, I think you have to end the color using `{1}`.

BlackCapCoder: I make it two lines instead of one to adhere to the 80 chars / column rule that seems to be used in the rest of the document.

Razesdark:
Thanks for the quick PR!

PR #856, opened 2018-07-01
Chanid search

Razesdark:
This is the #840 PR.

I had some stray commits to develop, cause i forgot to branch earlier so it got closed by github when I pushed a new version of develop to my forked repo


Razesdark:
If #859 gets merged, this will need to update its `@command` definitions.

Issue #820, opened 2018-04-01
[minor issue] handling of disabled comments.


pirate486743186:
This is a minor issue. Disabled comments could be handled more gracefully. When comments are disabled, it gives:

Error fetching data. Possible network issue.
Youtube Error 403: The video identified by the <code><a href="/youtube/v3/docs/commentThreads/list#videoId">videoId</a></code> parameter has disabled comments.

ritiek:
> The video identified by the `videoId` parameter has disabled comments.

I thinks its alright the current way it is. We can play the video just fine and this error occurs only when requesting for comments.

Razesdark:
I've added a PR for a small fix. It just removes the random HTML code which is passed from the API

PR #858, opened 2018-07-01
Filter out html code from pafy.util.GdataError

Razesdark:
References #820

pafys GDataError uses error messages from returned
api calls.

If comments were called on a video with comments disabled;
the error returned would be somewhat non-sensical to an
end user.

The error message contained some superflous information including
the html code pasted below.
`<code><a
href=\"/youtube/v3/docs/commentThreads/list#videoId\">videoId</a></code>`

This commit removes the superflous information and
has the function return a simple `Youtube error 403: The video has
comments disabled`

### DIFF
--- a/mps_youtube/commands/misc.py
+++ b/mps_youtube/commands/misc.py
@@ -107,7 +107,6 @@ def fetch_comments(item):
     try:
         jsdata = pafy.call_gdata('commentThreads', qs)
     except pafy.util.GdataError as e:
-        print(" " * util.getxy()[0] + "\r") # blank the li
         raise pafy.util.GdataError(str(e).replace(" identified by the <code><a href=\"/youtube/v3/docs/commentThreads/list#videoId\">videoId</a></code> parameter", ""))
     coms = [x.get('snippet', {}) for x in jsdata.get('items', [])]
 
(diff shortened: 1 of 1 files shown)


ritiek:
IMO it would be nicer to fix this in pafy itself.

Razesdark:
I disagree.

The current output is consistent with the error message we get from youtube's api, and becaue pafy is a generic wrapper for youtube's api, I don't think that it should alter the responses in any way. It is up to the projects utilizing it to handle the response.

mps_youtube/commands/misc.py
```diff
@@ -104,8 +103,13 @@ def fetch_comments(item):
...
 
-    jsdata = pafy.call_gdata('commentThreads', qs)
-
+    jsdata = None
+    try:
+        jsdata = pafy.call_gdata('commentThreads', qs)
+    except pafy.util.GdataError as e:
+        print(" " * util.getxy()[0] + "\r") # blank the li
+        print(str(e).replace(" identified by the <code><a href=\"/youtube/v3/docs/commentThreads/list#videoId\">videoId</a></code> parameter", ""))
+        return None
```
ids1024: Instead of printing and returning, this should probably re-raise the exception with the new text.

Issue #801, opened 2018-03-06
Consider better way to specify commands


ids1024:
The `@command` decorator currently uses a regex; this is fairly cryptic, somewhat error prone, and could be improved.

#789 added a list of command names, which should be unnecessary.

PR #859, opened 2018-07-01
Better @command definition that can register tab-completion strings

Razesdark:
This PR moves the list of commands from a static list in `util.py` to an argument
that is passed to the @command function. This way we dont have a separate list of tab completion.

ref: #801

### DIFF
--- a/mps_youtube/commands/__init__.py
+++ b/mps_youtube/commands/__init__.py
@@ -13,14 +13,13 @@ PL = r'\S*?((?:RD|PL|LL)[-_0-9a-zA-Z]+)\s*'
 
 ## @command decorator
 ##
-## The @command decorator takes two arguments, one is a regex that
-## is used to match the text command. All following arguments 
-## are strings that match the regex that is going to be added to
-## tab completion. 
+## The @command decorator takes a single regex followed by one or more
+## strings that corresponds to command names that will be added to
+## the tab completion.
 ##
 ## If your command has short-forms, only register the longer
 ## forms.
-## If you use several functions for the same command but different
+## If you use several functions and regexes for the same command but different
 ## arguments, append the completion string on EACH function, not only
 ## the first time you register it.
 def command(regex, *commands):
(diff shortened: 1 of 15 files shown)


mps_youtube/commands/__init__.py
```diff
@@ -11,10 +11,22 @@
 RS = r'(?:(?:repeat|shuffle|-[avfw])\s*)'
 PL = r'\S*?((?:RD|PL|LL)[-_0-9a-zA-Z]+)\s*'
 
-
-def command(regex):
+## @command decorator
+##
+## The @command decorator takes two arguments, one is a regex that
```
ids1024: It's not very important, but I wouldn't describe a polymorphic function as taking "two arguments"

mps_youtube/player.py
```diff
@@ -73,6 +73,7 @@ def play(self, songlist, shuffle=False, repeat=False, override=False):
                                                          override=self.override,
                                                          softrepeat=self.softrepeat)
                 self._playsong()
+                self.song_no += 1
```
ids1024: Uh, the changes in this file seem unrelated to the purpose of the commit or PR

ids1024:
This may not the the ideal solution, but it's definitely an improvement.

It might be worth thinking about something that results in less redundancy. Not using regexes at all seems like one option, but I don't know how that could be avoided given all the things it's used for. But many of the commands follow the same format of \<command-name\> {\<argument\>}, so perhaps it could be simplified in some way.

mps_youtube/player.py
vn-ki: You should not do this. This is handled by `CmdPlayer.next`

Razesdark: You're right, this is unrelated. It should not be here. I'll remove it.

mps_youtube/player.py
```diff
@@ -89,10 +89,10 @@ def play(self, songlist, shuffle=False, repeat=False, override=False):
             if config.SET_TITLE.get:
                 util.set_window_title("mpsyt")
 
+            print("songlistlen " + str(self.song_no) + ":" + str(len(songlist)))
```
ids1024: Still have this line (which presumably is debugging code not meant to be committed) and another change in this file.

mps_youtube/commands/__init__.py
Razesdark: Yeah, that was kind of embarrassingly bad.
I've updated it.

Issue #168, opened 2014-12-03
[Feature Request] Add song to mpd/mpc


lionawurscht:
Hey,
I just discovered your programm and love it. Unfortunately I also started using mpd a short while ago. Mpd can also play streams so it would theoretically possible to implement the option of adding the songlist to mpd (eg via mpc add) so they will be played later. This would allow me to play both local files and youtube songs in the same playing queue.

ids1024:
Players other than mpv and mplayer are not supported, but this works for me:

    set player mpc
    set playerargs add

ata2001:
> Players other than mpv and mplayer are not supported
Would you accept it, if someone implemented mpc like the other players?

ids1024:
> Would you accept it, if someone implemented mpc like the other players?

Yes, that would be reasonable. But `player.py` is currently a bit of a mess; it would probably be better to first refactor it to be class-based (a `Player` class, with `Mpv`, `Mplayer`, etc. classes derived from it).

ata2001:
Should we create a new issue then?

BobSpams:
The "mpc add" isn't working for me
I have mpc, mpd installed

PR #860, opened 2018-07-02
Mpd

### LINKED ISSUE #495: [Feature Request] Add 'headers' to a song added to mpd/mpc

vn-ki:
Closes #495 and #168 

This is how ncmpcpp looks while populating a generic soft pop yt playlist. mpsyt might work as a very good mpd client. We should add look into adding more mpd support.

2018-07-02-213016_800x300_scrot

### DIFF
--- a/mps_youtube/player.py
+++ b/mps_youtube/player.py
@@ -9,6 +9,7 @@ import subprocess
 import socket
 from urllib.error import HTTPError, URLError
 from abc import ABCMeta, abstractmethod
+from collections import OrderedDict
 
 
 from . import g, screen, c, streams, history, content, config, util
@@ -39,7 +40,7 @@ class BasePlayer:
             paused = True
         g.mprisctl.send(('pause', paused))
 
-    def status(self):
+    def init_status(self):
         msg = "%splayer%s set to %s%s%s" % (c.g, c.w, c.g, self.player, c.w)
         return dict(valid=True, message=msg, value=self.player)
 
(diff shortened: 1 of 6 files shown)


mps_youtube/player.py
```diff
@@ -138,18 +146,24 @@ def _launch_player(self):
...
+        if metadata is not None:
+            ret_meta['title'] = metadata['track_title']
+            ret_meta['arturl'] = metadata['album_art_url']
+            ret_meta['artists'] = [metadata['artist']]
+            ret_meta['album'] = metadata['album']
+
+        return ret_meta
+
+    def send_metadata_mpris(self):
+        metadata = tuple((v for k,v in self.generate_metadata().items()))
```
ids1024: You can use the `values()` method here: `metadata = tuple(self.generate_metadata().values())`

As a side note, you can leave off the extra set of parentheses with generator comprehensions like this: `tuple(v for k,v in self.generate_metadata().items())`. But with `.values()` you don't even need the generator comprehension.

ids1024: Although on second thought, can't you send the dict to the mpris process? Or a `collections.namedtuple` might be neater. It's probably best to use the same type to store this data in both places instead of converting to another format.

RaisonBlue:
Hello there ! Is this issue being reviewed ?

fauno:
FWIW I've rebased it to latest develop and it works :)

vn-ki:
I don't use mpd anymore, so I'm closing this PR. I'll be happy to review it anyone wants to pick up on this.

Phantom-Rehan:
@vn-ki How does this work, and what changes would I have to make to my mpd.conf and/or ncmpcpp config?

BobSpams:
Can someone try to create a new branch for this? I would love to be able to use mpsyt as a music backend

PR #863, opened 2018-07-07
Fix broken url's in readme

ferrybig:
Still broken without me finding proper replacements:

* https://github.com/np1/mps


Razesdark:
Thanks for helping out!

Issue #793, opened 2018-02-25
[Privacy issue] Unencrypted metadata sent to third party (Last.fm) by default


cpupower:
For every audio/video played using mpsyt, there is an **unencrypted** HTTP request sent to Last.fm including the title of the Youtube video. This is **by default** and it seems that there is no setting to disable it either.

In my opinion, this should not be the default setting, as it is not made clear, that even without specifiying any Last.fm credentials, personal data (i.e. metadata about the currently watched video/audio) is sent to a *third party* (i.e. Last.fm). Even worse the request is done in an unencrypted fashion and thus not only Last.fm, but also everyone else interested is able to get that metadata.

Please consider changing the default setting, to not acquire metadata from third parties, and never in an unencrypted fashion.

cpupower:
Thanks for implementing this, and at least providing an opt-out functionality.

Please note that pafy, a library which is a dependency of mps-youtube, also has a similar issue (HTTP instead of HTTPS). Without this also getting fixed, there is still data getting leaked (i.e. the YouTube video IDs). See also https://github.com/mps-youtube/pafy/issues/201.

Razesdark:
Thanks, I don't often look at the pafy repo so I didn't really see this. I will look into it

PR #867, opened 2018-07-12
Use https and add optout for metadata via last.fm

Razesdark:
As noted in #793, we were using `http` over `https` when looking up metadata, this pr changes it to use https instead.

It also adds yet another setting to disable metadata querying all together for those that have issues with this.

Metadata searching is on by default, but can be disabled by running `set lookup_metadata false`

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -331,6 +331,7 @@ class _Config:
             ConfigItem("window_size", "",
                 check_fn=check_win_size, require_known_player=True),
             ConfigItem("download_command", ''),
+            ConfigItem("lookup_metadata", True),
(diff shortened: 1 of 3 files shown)


mps_youtube/player.py
```diff
@@ -138,7 +138,7 @@ def _launch_player(self):
         pass
 
     def send_metadata_mpris(self):
-        metadata = util._get_metadata(self.song.title)
+        metadata = util._get_metadata(self.song.title) if config.LOOKUP_METADATA.get == True else None
```
ids1024: It would be more idiomatic to leave out the ` == True` when testing a boolean condition.

PR #870, opened 2018-07-28
Fix edge cases of player app setup

Eskuero:
Allow VLC to be autodetected and allow changing player if the old one was uninstalled.

### DIFF
--- a/mps_youtube/init.py
+++ b/mps_youtube/init.py
@@ -39,12 +39,15 @@ def init():
 
     # set player to mpv or mplayer if found, otherwise unset
     suffix = ".exe" if mswin else ""
-    mplayer, mpv = "mplayer" + suffix, "mpv" + suffix
+    mplayer, mpv, vlc = "mplayer" + suffix, "mpv" + suffix, "vlc" + suffix
 
     if not os.path.exists(g.CFFILE):
 
         if has_exefile(mpv):
             config.PLAYER.set(mpv)
+
+        elif has_exefile(vlc):
+            config.PLAYER.set(vlc)
 
         elif has_exefile(mplayer):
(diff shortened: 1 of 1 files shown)


Razesdark:
LGTM

mps_youtube/init.py
```diff
@@ -39,21 +39,29 @@ def init():
...
         config.save()
 
     else:
         config.load()
-        assign_player(config.PLAYER.get)  # Player is not assigned when config is loaded
+        try:
+            assign_player(config.PLAYER.get)  # Player is not assigned when config is loaded
+        except FileNotFoundError:
+            print("Configured player " + config.PLAYER.get + " does not exist.")
+            print("A new one must be specified by the command 'set player <mplayer|mpv|vlc>' within the app")
```
vn-ki: `set player` is not limited to mplayer, mpv and vlc.

Eskuero:
What if I put the elif of vlc on the last spot to hold it as last resort?

Which other players should I put on the example command?

Eskuero:
Any updates on this? Is it gonna get merged, what other changes should I do?

tommysolsen:
I just came across a case where I was installing mpsyt on a new computer without mplayer and having the program crash before being able to configure mpv as my player of choice. Having it autodetect players will be great.

I don't really use VLC, but if its not working properly I think we should only autodetect the two actually supported players, `mpv` and `mplayer`. 

I would like it to say something along the lines of `Install and set the player using _set player mpv|mplayer` instead of the current message. Because the reason the user sees the message in the first place is because mpsyt couldnt find any of them.

And while we do technically make it possible to use other players than those two, we don't support them officially so there is no need to mention that here.

PR #875, opened 2018-08-04
Add Arch Linux install instructions

murnux:
Similar to Ubuntu, add terminal command to install from the Arch Linux official repos.


ids1024:
Might as well include this. Thanks!

PR #892, opened 2018-10-03
[Suggestion] Show QRCode in video information panel

andreparames:
This is just a suggestion of a functionality I use on my private branch. I use NewPipe on my phone, and I often want to keep listening to the same video when I leave my laptop.

This PR shows a QRCode of the video URL in the video information panel (`i [number]`) to make it easier to open it on a mobile device, if the new option `show_qrcode` is enabled and the `qrcode` package is installed.

Suggestions and criticism are welcome!

screenshot

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -345,6 +345,7 @@ class _Config:
             ConfigItem("autoplay", False),
             ConfigItem("set_title", True),
             ConfigItem("mpris", not mswin),
+            ConfigItem("show_qrcode", False),
             ]
 
     def __getitem__(self, key):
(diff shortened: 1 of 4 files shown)


mps_youtube/config.py
```diff
@@ -345,6 +345,7 @@ class _Config:
             ConfigItem("autoplay", False),
             ConfigItem("set_title", True),
             ConfigItem("mpris", not mswin),
+            ConfigItem("show_qrcode", True),
```
vn-ki: I think the default value should be `False`. Any arguments to convince me it should be true?

andreparames: I considered making it `False`; my concern is that the feature is not very discoverable. Then again, I guess other features are as well.

andreparames: I agree :slightly_smiling_face: I've changed it!

andreparames: (Thanks for your help, btw!)

vn-ki:
Thank you for the contribution.

PR #923, opened 2018-12-24
fixing overwrite when audio and video for muxing are webm

### LINKED ISSUE #920: [bug] downloading separate video and audio webm stream overwrite each other.
when you use the download  command.
choose a webm video with out sound
for example for https://youtube.com/watch?v=p7XW6XUORfI
video 6 (webm)
and then a webm sound stream for merger letter.
audio 1 (webm)

Both have the exact same file name, and the video will overwrite the audio.
This happens for webm only.

ghost:
fix #920

Well, one way to fix this...

### DIFF
--- a/mps_youtube/commands/download.py
+++ b/mps_youtube/commands/download.py
@@ -66,7 +66,7 @@ def download(dltype, num):
             if url_au and ext_au:
                 # downloading video and audio stream for muxing
                 audio = False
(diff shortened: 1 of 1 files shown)


pirate486743186:
i just shoved it in to one line, this way, no more tmp variable....
`filename_au = _make_fname(song, ext_au) + "." + "audio"`

PR #924, opened 2019-01-05
fix crash disabled playback on other websites

pirate486743186:
fix crash with error "Watch this video on YouTube. Playback on other websites has been disabled by the video owner."

Example: "https://www.youtube.com/watch?v=NTUllANqqds"

also if searched for it ".Shinmai" (appears as stream 1)
The error appears once (no crash).
I guess that's intentional....

EDIT: This is actually an age restricted video... Not all age restricted videos are blocked this way though... Example: https://www.youtube.com/watch?v=3iMw1-m_V6U


pirate486743186:
ok... It seams the dev branch doesn't crash here....

Issue #930, opened 2019-01-11
Config option to disable history


chiraag-nataraj:
## Issue / Suggestion

Could we get an option to disable watch history? I know we already have `history clear`, but if you're not going to use the watch history at all, then it doesn't make sense to store it in the first place.

PR #934, opened 2019-01-17
Option to turn on and off history recording

vikramkashyap:
This commit adds an option not have watch and input history recorded, as requested in issue #930 .

I have created two configuration variables: "history", which decides whether watch history is recorded, and "input_history", which decides whether command input history is recorded. Information about these has been added to the help text. I also added a message that informs the user that history is off if they use the "history" command when history is turned off.

I've been looking through the python code and deciphering the inner-workings of mps-youtube. Hopefully I made the changes correctly, but please do tell if I missed something.

### DIFF
--- a/mps_youtube/commands/misc.py
+++ b/mps_youtube/commands/misc.py
@@ -48,7 +48,7 @@ def show_help(choice):
 @command(r'(?:q|quit|exit)', 'quit', 'exit')
 def quits(showlogo=True):
     """ Exit the program. """
-    if has_readline:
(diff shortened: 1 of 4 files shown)


pirate486743186:
you normally start a new brach based on the current main branch, when proposing a change. You instead put both changes in the develop branch of your fork...

It should still be ok though.

vikramkashyap:
I have removed the other change. It is on its own branch called "optional_history" and has only one commit to be merged now.

Issue #935, opened 2019-01-18
Update Dockerfile to current best practices


willhallonline:
## Issue / Suggestion
Current ```Dockerfile``` does not follow best practices.

1. Maintainer deprecated (replaced with layers) https://docs.docker.com/engine/reference/builder/
2. Layers required for inspection and management
3. Reduce overall use of layers where possible
4. Fix Dockerfile ```FROM``` to a version (latest not a secure option).

willhallonline:
Fixed by #936

willhallonline:
Done. :+1:

PR #936, opened 2019-01-18
Dockerfile improvements

willhallonline:
1. Drop Maintainer (deprecated)
2. Add layers for better understanding of container build
3. Reduce layer use for easier caching.
4. Fix Dockefile ```FROM``` image to numbered version and OS (3-stretch)


willhallonline:
Fixes #935

PR #963, opened 2019-03-21
Synchronize playlist prefixes with pafy

lopsided98:
pafy has [support for a couple of playlist prefixes](https://github.com/mps-youtube/pafy/blob/develop/pafy/playlist.py#L23) that are not recognized by mps-youtube, so this PR adds them to the `pl` command regex. Note that OL prefixes are actually not supported by pafy without https://github.com/mps-youtube/pafy/pull/226.


tommysolsen:
Thank you!

Issue #977, opened 2019-05-06
Getting the Youtube Error 403!


CharlesCage:
## Issue / Suggestion

Receiving the following error:
Youtube Error 403: The request cannot be completed because you have exceeded your <a href="/youtube/v3/getting-started#quota">quota</a>.

ritiek:
It seems like user base has grown more than a single API key can handle.

You could generate your own YouTube Data API key (Guide: https://elfsight.com/help/how-to-get-youtube-api-key/) and then set this key in mpsyt with:
```
> set api_key AIzaSyC.......
```

It should work now.

lightonflux:
If the amount of requests for that key is a problem, why can you play the next song without issue?

ritiek:
> If the amount of requests for that key is a problem, why can you play the next song without issue?

@lightonflux I just tried and it seems like there are 2 different causes for the 403 error.

The first one is the one reported in this issue:
> Youtube Error 403: The request cannot be completed because you have exceeded your quota.

which is caused due to making more API requests in a day than the API key allows for.

The second error (the one I think you're talking about) which I receive is simply:
> Problem playing last item: HTTP Error 403: Forbidden

To me, this seems like a bug with the YouTube API itself. Pick up any YouTube video and get its audio URL with youtube-dl and like 1/5 times it would result in a forbidden URL. Such as:

```
$ curl --head $(youtube-dl -x https://youtube.com/watch?v=p_vSh1DUsD4 -g)
```

And it looks like mpsyt caches these audio URLs so if you get this forbidden error on a track once, you will always get this same error whenever you try to play that same track through mpsyt. Until you delete the URL cache file in `~/.config/mps-youtube/cache_py_3.*` and try your luck again.

ritiek:
What I think we could do to workaround this error would be test beforehand if the generated audio URL gives a 403 forbidden. If it does, keep retrying.

PatOConnor43:
I can confirm that deleting `~/.config/mps-youtube/cache_py_3.*` solved the issue for me. Thanks for the workaround!

eskiq:
I'm heaving problem "https: HTTP error 403 Forbidden" on some videos while using terminal and command mpv m.m3u, but never had any problem playing same list outside terminal. Using terminal I still can go back and load something that was not loaded in the first place, while mps will not load it at all. There is 1 of 3 commands that I use in my mpv.conf depending if I wanna watch with or without video and they gave a little different results what will play in terminal and what not.
[...]

PR #979, opened 2019-05-19
added api_key guide to mitigate #977

gusals3587:
Some non-advanced users might be confused with the occasional 403 errors mps-youtube receives, I'm hoping to mitigate this by adding a guide to add their own custom api_key


gusals3587:
whoops, forgot there was already a FAQ section, my bad, will contribute there

PR #998, opened 2019-07-15
use json format for config file

dt-rush:
This PR replaces the config file format with human-friendly JSON, and also adds code to seamlessly convert old configs to JSON if detected.

---

## rationale

As the recent API key overload made clear (issue #551), users being able to edit the config themselves is an important feature (modifying the configured default API key is not really feasible by hand, as one user remarked).

The config is currently stored as a pickled binary file, which is not user-friendly.

Pickled binary storage is not necessary for any reasonable disk-space reasons. The difference in the old config style vs json is on the order of tens of bytes, as can be seen:

```
pi@raspberrypi:~/.config/mps-youtube $ python3

>>> import pickle
>>> import json
>>> with open('config', 'rb') as cf:
...     with open('config.json', 'w') as cfj:
...             json.dump(pickle.load(cf), cfj)
... 
>>> 
KeyboardInterrupt

pi@raspberrypi:~/.config/mps-youtube $ ls -l
total 20
-rw-r--r-- 1 pi pi  708 Jul 15 03:14 config
-rw-r--r-- 1 pi pi  676 Jul 15 03:47 config.json
```

Signed-off-by: dt-rush <nickp@balena.io>

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -3,6 +3,7 @@ import re
 import sys
 import copy
 import pickle
+import json
 from urllib.request import urlopen
 from urllib.error import HTTPError
 from urllib.parse import urlencode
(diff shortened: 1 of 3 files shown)


ids1024:
Thanks! The `set` command exists to modify configuration, and I don't see a strongly compelling reason for the format to be text based. But generally text based configuration formats are nicer to have, and there isn't much reason not to.

TsukiZombina:
> Thanks! The `set` command exists to modify configuration, and I don't see a strongly compelling reason for the format to be text based. But generally text based configuration formats are nicer to have, and there isn't much reason not to.

The `set` command doesn't work for me, just changing configuration file, as I mention in #551.

PR #1001, opened 2019-07-26
Misalignment due to East East Asian chars fix

Nojus297:
Due to East East Asian chars being more than two columns wide on terminal, video list often was misaligned and looked like this:

mps_yt_1
(Whole table doesn't even fit into the terminal window)

Added `util.correct_truncate` method which truncates a string taking into account East Asian width chars. `util.`ea_pad' now pads correctly.
Also `content.generate_songlist_display` and `ListView.content` functions now use custom string formatting instead of the default one and output correct table:

mps_yt_2

**Note:** the `util.real_len` function doesn't take into account Unicode emojis. Because of that, the table is still a bit misaligned (line 12 & 20). 
I suggest to let [wcwidth](https://github.com/jquast/wcwidth) in `real_len` function take care of it, it should solve the problem. If the additional dependency is okay.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ pafy
 vi.py
 MANIFEST
 .dev
+.vscode
 # Packages
 *.egg
 *.egg-info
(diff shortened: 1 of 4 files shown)


Nojus297:
Any chance to get this at least reviewed?

tommysolsen:
Thanks for this PR, Nojus!

Nojus297:
I'm glad that it got merged.
What do you think about using [wcwidth](https://github.com/jquast/wcwidth) module?

PR #1010, opened 2019-09-12
Fix phrasing and typos in README.rst

joshnatis:


Nojus297:
I think you misunderstood the meaning of this paragraph. It talks about two different `youtube-dl` versions (one is from the official website download link and the other one is from linux repositories). Apparently neither of these two versions are compatible with `mps-youtube`. Because of that, it's recomending to install (and upgrade) using `pip3`.

joshnatis:
You're right, thank you for catching that. 

I think that paragraph is a little ambiguous -- for others' future reference could you update it with the explanation you provided for me?

Thanks

Nojus297:
Sorry for the late reply.
I'm not an English speaker, so I'm not that good with wording, so I don't think I could make a decent PR with an explanation.
You could open a new PR I guess, it might get merged.
Though it seems that stuff happens rather slowly here.

PR #1028, opened 2019-10-25
Upgrade parfy on Docker  in order to avoid `'NoneType' object has no attribute 'group`  bug

vtrbtf:


hrnr:
Why is this problem? Shouldn't we get the latest version as a dependency of mps-youtube?

Issue #486, opened 2016-05-11
Audio not encoding


aeryx:
I've installed ffmpeg with mp3lame, configured mpsyt to encode in mp3 256k and edited the text config file to point to ffmpeg binary, but all dowloads are still saved with the source encoding. What am I missing?

speedemonV12:
I am also experiencing this problem. Both on Arch and Ubuntu. Any help appreciated. Thanks!

speedemonV12:
Update: Encoder seems to work when downloading the m4a source. That seems to be the only one that mps automatically converts to mp3. The rest all download in their source encoding and not being converted to mp3

spcmd:
@speedemonV12 I had to add webm to the "valid for:" lines in the transcode file. See [this comment](https://github.com/mps-youtube/mps-youtube/issues/389#issuecomment-164239705) from #389

ZaxonXP:
Indeed, when the source is m4a AND the "download_command" is set to "none" then the conversion take place. Otherwise the file is not converted. It would be nice if the conversion is also done when the external downloader is in use.

PR #1032, opened 2019-11-04
Fix issue #486: enabling conversion when download_command is set

mateuszr:
Fixing issue #486 when file conversion won't take place if "download_command" is set.

### DIFF
--- a/mps_youtube/commands/download.py
+++ b/mps_youtube/commands/download.py
@@ -332,6 +332,21 @@ def external_download(song, filename, url):
     util.dbg("Downloading using: %s", " ".join(cmd_list))
     subprocess.call(cmd_list)
 
+def _transcode(song, filename, audio=False, allow_transcode=True):
+    """ Transcode filename
+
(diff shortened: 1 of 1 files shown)


mps_youtube/commands/download.py
```diff
@@ -332,6 +332,21 @@ def list_string_sub(orig, repl, lst):
     util.dbg("Downloading using: %s", " ".join(cmd_list))
     subprocess.call(cmd_list)
 
+def _transcode(song, filename, audio=False, allow_transcode=True):
```
hrnr: taking `allow_transcode` in transcode function does not make sense. Please remove this, we don't pass it anyway.

hrnr:
Thanks, looks good to me, except a small issue.

mps_youtube/commands/download.py
mateuszr: fixed in the next commit

mateuszr:
The comment from @hrnr has been addressed.

Issue #964, opened 2019-03-31
Read `mpv.conf` for mpv global settings


anishmittal2020:
## Your Environment

mpsyt version      : 0.2.8
   notes           : released 17 February 2018
pafy version       : 0.5.4 (youtube-dl backend)
youtube-dl version : 2019.01.17
Python version     : 3.7.3 (default, Mar 30 2019, 15:09:56) 
[GCC 5.4.0 20160609]
Processor          : x86_64
Machine type       : x86_64
Architecture       : 64bit, ELF
Platform           : Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid
sys.stdout.enc     : UTF-8
default enc        : utf-8
Config dir         : /home/nikhil/.config/mps-youtube
env:TERM           : xterm-256color
env:SHELL          : /bin/bash
env:LANG           : en_US.UTF-8
env:LANGUAGE       : en_US:en

Ubuntu 18.04

MPV

## Issue / Suggestion

- With the following mpv settings:

```bash
cat ~/.config/mpv/mpv.conf 
ytdl-raw-options=sub-format=en,write-srt=
```

- I am able to play Youtube videos with subtitles. For example:

```bash
mpv https://www.youtube.com/watch?v=oPPJ7GGDyAw
```

Also I have symlinked the files inside `~/.config/mps-youtube`: 

```bash
lrwxrwxrwx 1 nikhil   17 Mar 31 20:03 mpv-input.conf -> ../mpv/input.conf
lrwxrwxrwx 1 nikhil   15 Mar 31 20:03 mpv-mpv.conf -> ../mpv/mpv.conf
```

- But `mpsyt` doesn't play the video with subtitle. Am I missing out some detail? Please suggest, as there is no documentation on how to do this.

Please add this feature if it is not supported.

Issue #331, opened 2015-07-28
Subtitle support


vbmvbm:
It would be nice if mpsyt could support YouTube subtitles.

I've spent a few hours on it and found out that it is quite easy to fetch the subtitle list, the text and convert it to .srt format (which can be opened by players like mpv or mplayer), as shown in the following sketch of code I've just written:

```python
#!/usr/bin/env python3

# used to fetch the content from a url
import urllib.request as ur

# used to parse the xml
import xml.etree.ElementTree as et

# used to unescape the text, e.g., &lt; to >
import html

# used to get the temp dir
import tempfile

# dictionary which keeps the order of insertion
from collections import OrderedDict as od


# Receives a time in seconds
# Returns its representation in the srt format hh:mm:ss:MMM
def time_to_str(time):
    h = int(time/3600)
    m = int((time % 3600) / 60)
    s = int(time % 60)
    ms = int((1000 * time) % 1000)
    return '%02d:%02d:%02d,%03d' % (h, m, s, ms)

# Receives a list of tuples in the form (start, end, text)
# Returns a string that can be saved to a valid .srt file
def track_to_srt(texts):
    i = 1
    srt = ''
    events = {}
    for text in texts:
        start, end = text[0], text[0] + text[1]
        if start in events:
            start += 0.00001 # prevent conflict
        events[start] = html.unescape(text[2])
        events[end] = None
    # This function may seem too much complicated, but I think this is needed 
    #  to deal with live closed-captions with superposition.
    text = []
    i = 1
    start = 0
    for t in sorted(events):
        # This will deal with live closed-captions to prevent hiding an old line
        #  before showing a new one, if these events are too close.
        # This doesn't affect normal subtitles, because in those there is no
        #  superposition, therefore events[t] will always be None when text is non
        #  empty, and the following variable will always be true.
        update = not (text and events[t] is not None and t - start < 0.15)
        # if there is some text, put it into the srt
        if text and update:
            srt += str(i)
            srt += '\n' + time_to_str(start) + ' --> ' + time_to_str(t)
            srt += '\n' + '\n'.join(text) + '\n\n'
            i += 1
        # change
        if events[t] is None and text: # removing
            text.pop(0)
        else: # adding
            text.append(events[t])
        if update:
            start = t
    # If there were only normal subtitles, the following would be enough instead of
    #  this entire function:
    '''
    for text in texts:
        srt += str(i)
[...]

PR #1038, opened 2019-12-06
Adding subtitle support via workaround

yc-tsui:
As noted in #331 and #964, there are no subtitles when opening videos with mpv through mpsyt. Since `mpv [link]` does show subtitles, the problem is likely in how mpsyt calls mpv, but I'm not sure exactly where. A temporary workaround is to add a new command, similar to the 'copy link to clipboard', except that the link goes to opening mpv directly. In the video list, typing `m <number>` will do `mpv [link]` and subtitles will now work.

This isn't the ideal solution (it imports `os`, check for whether mpv is installed is not implemented, no mplayer support), but there's hasn't been any progress on adding subtitle support, so this workaround might be useful for people at the meantime

### DIFF
--- a/mps_youtube/commands/misc.py
+++ b/mps_youtube/commands/misc.py
@@ -5,6 +5,7 @@ import traceback
 from urllib.request import urlopen
 from urllib.error import HTTPError, URLError
 from .. import player
+import os
 
 try:
     # pylint: disable=F0401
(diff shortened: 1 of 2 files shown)


tommysolsen:
If just passing the video url is enough to have it display subtitles, i guess either the way the video url is resolved and passed to the MPV player class or some flag we're setting that is hindering them from being shown correctly.

This would definitely be something that is best off being a part of the core video watching flow, as the extra functionality seemingly often gets overlooked, though if no one steps up and is willing to submit a PR fixing the issue in the main mpv player, then I'dd be happy to merge this.

Gonna leave it for a short while and see

PR #1065, opened 2020-02-28
Transcode to mp3 from webm too

Atcold:
Currently `webm` are not transcoded to mp3.
This fixes it.


tommysolsen:
Thank you for your contribution!

PR #1072, opened 2020-04-03
move pafy.set_api_key before cache.load()

tchernomax:
fix https://github.com/mps-youtube/mps-youtube/issues/1070


tommysolsen:
Thanks for your contribution!

PR #1088, opened 2020-05-16
Fixed ValueError: time data does not match format

francutura:

PR #1096, opened 2020-06-13
Try to parse the date strings using several formats.

eldipa:
Youtube seems that changed the format of the date times so this commit
fixes that. In case of a regression (from Youtube), we try to parse the
string using several formats and not just one.

Refactor the code so `yt_datetime` and `yt_datetime_local` use the same
code.

See issue #1088


iamtalhaasghar:
related to https://github.com/mps-youtube/yewtube/pull/1156

Issue #1101, opened 2020-07-11
Problem playing last item: unconverted data remains: Z


louisabraham:
## Issue 

When trying to display information about a search result with `i 1`, I get the following error:

    Problem playing last item: unconverted data remains: Z

## Your Environment

I'm on manjaro ARM and installed mpsyt from the github repository.

```
mpsyt version      : 0.2.8
   notes           : released 17 February 2018
pafy version       : 0.5.5 (youtube-dl backend)
youtube-dl version : 2020.06.16.1
Python version     : 3.8.3 (default, May 19 2020, 23:51:58) 
[GCC 9.3.0]
Processor          : 
Machine type       : aarch64
Architecture       : 64bit, ELF
Platform           : Linux-5.7.0-2-MANJARO-ARM-aarch64-with-glibc2.17
sys.stdout.enc     : utf-8
default enc        : utf-8
Config dir         : /home/louisabraham/.config/mps-youtube
env:TERM           : xterm-256color
env:SHELL          : /bin/bash
env:LANG           : en_US.utf8
```

poyenc:
The root cause is that when `mpsyt` trying to parse video published date string, the employed function [`time.strptime()`](https://docs.python.org/3/library/time.html#time.strptime) did not recognize the ending `Z`, which is **time zone designator**, in data string. (defined by [ISO 8601](https://www.w3.org/TR/NOTE-datetime) according to [Youtube Data API v3](https://developers.google.com/youtube/v3/docs/videos#properties))

In function `commands.misc.info()` located in file _commands/misc.py_, we can simply add `Z` in the format string to avoid this error, but I think we need more graceful way to handle this issue.
```diff
   elif g.browse_mode == "normal":
        g.content = logo(c.b)
        screen.update()
        screen.writestatus("Fetching video metadata..")
        item = (g.model[int(num) - 1])
        streams.get(item)
        p = util.get_pafy(item)
-       pub = time.strptime(str(p.published), "%Y-%m-%d %H:%M:%S")
+       pub = time.strptime(str(p.published), "%Y-%m-%d %H:%M:%SZ")
        screen.writestatus("Fetched")
        out = c.ul + "Video Info" + c.w + "\n\n"
```

louisabraham:
I think it's ok to put a condition to test if the last character is Z (in case the API drops it). What do you imagine could be more graceful?

poyenc:
[...]

PR #1102, opened 2020-07-11
Fix issue #1101: ValueError raises when view video information

poyenc:
The `video_info()` employs [`strptime()`](https://docs.python.org/3/library/datetime.html?highlight=datetime#datetime.datetime.strptime) to parse date string, but it cannot handle the [ISO 8601](https://www.w3.org/TR/NOTE-datetime) **time zone designator** at the end of string. For now we assume there is always a 'Z' time zone designator thus add 'Z' in the `strptime()` format string.

fix issue: #1101


poyenc:
ref #1110

tommysolsen:
Thank you for your contribution.

Issue #1103, opened 2020-07-13
The MPRIS process error message is not very descriptive


poyenc:
## Issue
When I use [*PuTTY*](https://www.putty.org) (on _Windows_) to connect to an *Ubuntu 18.04* desktop machine (without X11 forwarding)
`mpsyt` shows following error message:

***old error message***
```
mpris interface couldn't be initialized. Is dbus properly configured?
```
I don't know how to fix it until I modified the `mpris.main()` function to let it print `DBusException.get_dbus_message()` return value, and it saves me a lot of time to seek a solution.

***new error message***
```
Unable to autolaunch a dbus-daemon without a $DISPLAY for X11
```

By setting the `$DISPLAY` environment variable to `:0`, I can now playing audio without any problems. And the play progress bar display normally too.

```bash
$ DISPLAY=:0 mpsyt playurl ioNng23DkIM
```

## Suggestion
Show the exact reason when MPRIS process exited abnormally.

*possible modification in mpris.py*
```diff
def main(connection):
    """
        runs mpris interface and listens for changes
        connection - pipe to communicate with this module
    """

    mprisctl = Mpris2Controller()
    try:
        mprisctl.acquire()
-    except dbus.exceptions.DBusException:
+    except dbus.exceptions.DBusException as e:
-        print('mpris interface couldn\'t be initialized. Is dbus properly configured?')
+        print('mpris interface couldn\'t be initialized. reason: {}'.format(e.get_dbus_message()))
        return
    mprisctl.run(connection)
    mprisctl.release()
```

## Your Environment
**OS version**: *Ubuntu 18.04*
**music player**: *(default)*
```
mpsyt version      : 0.2.7.1
   notes           : released 6 July 2016
pafy version       : 0.5.2 (youtube-dl backend)
youtube-dl version : 2020.06.16.1
Python version     : 3.6.9 (default, Apr 18 2020, 01:56:04)
[GCC 8.4.0]
Processor          : x86_64
Machine type       : x86_64
Architecture       : 64bit, ELF
Platform           : Linux-4.15.0-109-generic-x86_64-with-Ubuntu-18.04-bionic
sys.stdout.enc     : UTF-8
default enc        : utf-8
Config dir         : /home/poyenc/.config/mps-youtube
env:TERM           : screen-256color
env:SHELL          : /bin/bash
env:LANG           : en_US.UTF-8
```

PR #1104, opened 2020-07-14
Fix issue #1103: Make MPRIS process error message more descriptive

poyenc:
When MPRIS process exited abnormally, show the exact reason if we cannot successfully setup dbus.

fix issue: #1103


tommysolsen:
Looks good to me!

Thank you for your contribution

Issue #1134, opened 2020-12-17
KeyError: data


louisabraham:
## Issue / Suggestion

On any video:

```
Exception in thread Thread-2: Mb                                             
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/mpris.py", line 227, in bindmpv
    self.setproperty(resp['name'], resp['data'])
KeyError: 'data'
Traceback (most recent call last):
  File "/usr/bin/mpsyt", line 11, in <module>
    load_entry_point('mps-youtube==0.2.8', 'console_scripts', 'mpsyt')()
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/main.py", line 153, in main
    if matchfunction(i.function, i.regex, userinput):
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/main.py", line 70, in matchfunction
    func(*matches)
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/commands/play.py", line 102, in play
    g.PLAYER_OBJ.play(songlist, shuffle, repeat, override)
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/player.py", line 79, in play
    self._playsong()
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/player.py", line 138, in _playsong
    self._launch_player()
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/player.py", line 298, in _launch_player
    self.launch_player(cmd)
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/players/mpv.py", line 134, in launch_player
    self._player_status(self.songdata + "; ", self.song.length)
  File "/home/louisabraham/.local/lib/python3.8/site-packages/mps_youtube/players/mpv.py", line 187, in _player_status
    if resp['data'] is not None:
KeyError: 'data'
```

## Your Environment

```
mpsyt version      : 0.2.8
   notes           : released 17 February 2018
pafy version       : 0.5.5 (youtube-dl backend)
youtube-dl version : 2020.12.14
Python version     : 3.8.6 (default, Oct  6 2020, 22:56:58) 
[...]

Issue #1128, opened 2020-12-01
Mpv crashes mps-youtube, and MPlayer plays only 1mn. 


galgot:
Hi, as the title says, setting mpv as player and trying to read a track makes mpsyt quit with this error : 
```
Traceback (most recent call last):
  File "/usr/local/bin/mpsyt", line 8, in <module>
    sys.exit(main.main())
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/main.py", line 153, in main
    if matchfunction(i.function, i.regex, userinput):
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/main.py", line 70, in matchfunction
    func(*matches)
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/commands/play.py", line 102, in play
    g.PLAYER_OBJ.play(songlist, shuffle, repeat, override)
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/player.py", line 79, in play
    self._playsong()
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/player.py", line 138, in _playsong
    self._launch_player()
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/player.py", line 298, in _launch_player
    self.launch_player(cmd)
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/players/mpv.py", line 135, in launch_player
    self._player_status(self.songdata + "; ", self.song.length)
  File "/usr/local/lib/python3.8/site-packages/mps_youtube/players/mpv.py", line 188, in _player_status
    if resp['data'] is not None:
KeyError: 'data'
```

If setting player to MPlayer, it reads only 1 minute of the track.
I also get this msg before it reads : 
```
/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/subprocess.py:844: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stdout = io.open(c2pread, 'rb', bufsize)
```

This is on a MacBook Pro running Mac OS X 10.11.6. Here infos on the mpsyt install : 
mpsyt version      : 0.2.8
   notes           : released 17 February 2018
pafy version       : 0.5.5 (youtube-dl backend)
youtube-dl version : 2020.11.26
Python version     : 3.8.5 (default, Sep  9 2020, 23:31:58)
[Clang 8.0.0 (clang-800.0.42.1)]
Processor          : i386
[...]

PR #1135, opened 2020-12-17
fix #1134

louisabraham:

PR #1149, opened 2021-02-27
Fix issues  #1148 (and  #1128)

yaourtiere:
The change if bufsize=1 to bufsize=0 seems to work, I got no problem with that.


yaourtiere:
It also fix the issue #1134  (another patch is #1135 , which use the `dict.get()` method instead of an existence test)

iamtalhaasghar (approved):
Yes, this commit fixed the issue #1134

otisdog8:
Could you merge it into here?
https://github.com/otisdog8/termitube/
I'm making a fork because the repository is mostly unmaintained

PR #1153, opened 2021-03-12
fix help config

khonsaloh:
There is the following error on the help menu:
` set show_mplayer_keys true|false - show keyboard help for mplayer and mpv`
when setting this option with that command shows the following error:
`Unknown config item: show_mplayer_keys`
The actual command for this is:
`set show_player_keys true|false`


otisdog8:
Could you merge it into here?
https://github.com/otisdog8/termitube/
I'm making a fork because the repository is mostly unmaintained

henkmet:
I have made the pull request on your fork. For future reference, if you want to apply a PR from another repo, you can use 'git remote add <name> <url>' then 'git fetch <name>' and 'git cherry-pick <hash>'. See for example https://stackoverflow.com/questions/13788945/how-to-cherry-pick-from-a-remote-branch

PR #1203, opened 2023-01-27
merge yewtube codebase in mps-youtube

iamtalhaasghar:
This will enable mps-youtube to be used without Youtube API v3.


iamtalhaasghar:
Can you kindly review this pull request? @ids1024

mps_youtube/players/vlc.py
```diff
@@ -31,3 +36,13 @@ def launch_player(self, cmd):
 
     def _help(self, short=True):
         pass
+
+    def _kill_instance(self):
```
ids1024, line 40: For a process spawned from Python, one may want to use `Popen.terminate` rather than trying to use killall. That should be portable and not kill another vlc instance if it happened to be running.

PR #1050, opened 2020-01-08
Update Dockerfile and build dockerfile automatically using github actions

tommysolsen:
**Before merging this PR we need to set up some tokens for Github Actions to be able to push to dockerhub**

# Dockerfile changes
We now build the Docker image from the source code in the repository instead of downloading the newest version of pip. This way we can release multiple images of different (and unreleased versions) through dockerhub.

# New mpsyt organization on dockerhub.
I've created a new organization for mpsyt on dockerhub. I'll have to somehow get in touch with contributors in the mpsyt org in order to add them to the org and sort out owner rights at some point.

# Github actions
This PR also adds a few simple workflows that lets us automatically build images when we update the master and develop branch of the repository. The images are pushed to the aforementioned newly created org.

In order to facilitate that, we need to set a dockerhub secret to grant push rights. However it looks like only @ids1024 has access to do that. I can add him as owner to the dockerhub org, then he can create a security token and add that to the repo secrets and then we can automatically perform changes here.

We could/should also set up some schedule based workflows which would check and rebuild the images if youtube-dl has been updated. As well as building tagged versions of the software as well.

PR #1218, opened 2023-03-15
Dockerfile minor improvements

pataquets:
- Prevent `pip install` from persisting cache (improves size).
- Correctly set `DEBIAN_ENVIRONMENT` env variable. As of now, the env variable is not carried on next instructions. Using `ARG` sets it for all the remaining image build.

However, I've noticed that the Docker image is not build from source, but installed from PyPI repos, which is not the common usage pattern. That is, building from local cloned code, with the added benefit of being useful while developing.


iamtalhaasghar:
Search for existing docker issues and ping me. There's already one.

pataquets:
@iamtalhaasghar I've searched just before and found #1050, which I mistakenly took by an outdated issue, going just the opposite: just baking the binary.
After your comment, I've double-checked and there's even a comment by me pinging another user :facepalm:. My apologies. Pinging you from there.

PR #1261, opened 2023-11-19
Fix mix command

albino:
YouTube updates mean that `youtube-search-python` can no longer fetch mixes properly; they behave slightly differently to playlists.

The solution is to split these functions out and fetch the data using `yt-dlp`.


antonvice (approved):
looks good to me

Issue #1064, opened 2020-02-27
Incomplete search results


StylinGreymon:
Looking over a recent search, I realized that there were quite a few videos missing from the results, most notably a 1-month gap between January and February.
![image](https://user-images.githubusercontent.com/5328232/75424931-7ec10700-58f6-11ea-966f-f63086ab76d3.png)

Not sure when this started happening, and the full list of videos shows up regularly in a web browser.

## Your Environment
>mpsyt version      : 0.2.8
   notes           : released 17 February 2018
pafy version       : 0.5.5 (youtube-dl backend)
youtube-dl version : 2020.02.16
Python version     : 3.8.1 (default, Jan 22 2020, 06:38:00) 
[GCC 9.2.0]
Processor          : 
Machine type       : x86_64
Architecture       : 64bit, ELF
Platform           : Linux-5.4.18-1-MANJARO-x86_64-with-glibc2.2.5
sys.stdout.enc     : utf-8
default enc        : utf-8
Config dir         : /home/greymon/.config/mps-youtube
env:TERM           : screen-256color
env:SHELL          : /bin/zsh
env:LANG           : en_US.UTF-8

EDIT: I can't reproduce this error on any other search besides game grumps, but for some reason both 'user gamegrumps' and 'channels gamegrumps' return incomplete results, though the exact result differs.

UpHillSolutions-roberth:
I just recently started using yewtube so apologies for being so late however this issue is also happening for me. 

It looks like you are doing a `channels` or `user` search. 

I'm seeing this issue in both of those search query types as well. 

I have solved the issue on my end and submitted a pull request for the fix: https://github.com/mps-youtube/yewtube/pull/1282

I hope this helps!

PR #1282, opened 2024-08-21
Fix Issue #1064 and #1283: User and Channel searches to return all videos including optional filtering of search terms

UpHillSolutions-roberth:
Updated `pafy.py` to fix Issue #1064 
- Modified the `all_videos_from_channel` function to return all videos from a channel or user by looping through each page of the url playlist and then returning all videos. 

The previous method only returned the first page of results which meant that search queries for channels/users with more than 100 videos were only displaying the most recent 100 videos. This fix ensures that all videos are returned by using the same logic as `get_playlist`.

This fixes the following search queries to ensure all videos for the user/channel are returned (using videogamedunkey as an example but this applies to all users/channels): 
- `user videogamedunkey`
- `channels videogamedunkey 1`

### DIFF
--- a/mps_youtube/commands/search.py
+++ b/mps_youtube/commands/search.py
@@ -216,7 +216,14 @@ Use 'set search_music False' to show results not in the Music category.""" % ter
         else:
             failmsg = "User %s not found or has no videos."  % termuser[1]
     msg = str(msg).format(c.w, c.y, c.y, term, user)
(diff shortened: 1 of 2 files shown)


UpHillSolutions-roberth:
Updated `search.py` to fix Issue #1283. 

The search query for `user userName/searchTerm` now filters the output of `all_videos_from_channel` by the optional search term if provided otherwise all videos by the user are returned.

Issue #942, opened 2019-01-28
Video always play at 640x360


mrhieu059:
## Issue
Mps-youtube never play with quality higher than 640x360

## My Environment
Using Arch Linux with newest kernel (4.20.4)
https://pastebin.com/fJsE4Stp (mpsyt --version, briefly 0.2.8)
Mplayer using as default for mps-youtube
Mps's config showed max resolution 2160p
https://imgur.com/QQJPEad (My screenshot that show download option showed at max 1920x1080) but it only play at 640x360 :(

Carl-Robert:
Same issue on Windows 10 x64 LTSB and mpv player and mpsyt 0.28.
Using the pip version, because the precompiled .exe has outdated youtube-dl.

anishmittal2020:
I too am facing the same issue.

$ mpsyt --version
mpsyt version      : 0.2.8
   notes           : released 17 February 2018
pafy version       : 0.5.4 (youtube-dl backend)
youtube-dl version : 2019.01.17
Python version     : 3.7.3 (default, Mar 30 2019, 15:09:56) 
[GCC 5.4.0 20160609]
Processor          : x86_64
Machine type       : x86_64
Architecture       : 64bit, ELF
Platform           : Linux-4.15.0-46-generic-x86_64-with-debian-buster-sid
sys.stdout.enc     : UTF-8
default enc        : utf-8
Config dir         : /home/nikhil/.config/mps-youtube
env:TERM           : xterm-256color
env:SHELL          : /bin/bash
env:LANG           : en_US.UTF-8
env:LANGUAGE       : en_US:en

pirate486743186:
mps-youtube can only stream normal videos(video and audio). but Most of the available streams from youtube are actually video only or audio only. Often, 360 is the only one available and sometimes also 1280.

Mati20187:
> mps-youtube can only stream normal videos(video and audio). but Most of the available streams from youtube are actually video only or audio only. Often, 360 is the only one available and sometimes also 1280.

I don't think that's correct because "mpv + youtube_link" outputs videos in 720p

uncleungie:
I have the same issue - all streams play at a 640 max vertical resolution. Downloading the file yields a full quality version. Is there any way to stream videos at a higher resolution than SD?

sikosis:
Shame there was no fix for this ...

lamyergeier:
[...]

PR #1284, opened 2024-08-24
Implemented option to pipe url directly to mpv

thatonearchguy:
Hi!

Thanks for this project, it's been of great use on my ancient 2007 Macbook Pro - I can use mpv with --hwdec and --vo flags to get hardware accelerated video playback, which on a 17 year old CPU is a very welcome help.

However I was running into the 640x360 issue as in #942, so I decided to write this little patch to allow the URL to be piped directly to mpv and the --ytdl-format player args to be used to control the output instead (in my case limiting the resolution to 720p as the GPU runs out of VRAM)

Happy to make any changes needed (what default values would you suggest setting for the stream dict?)

### DIFF
--- a/mps_youtube/config.py
+++ b/mps_youtube/config.py
@@ -343,6 +343,7 @@ class _Config:
             ConfigItem("pages", 3, minval=1, maxval=100),
             ConfigItem("autoplay", False),
             ConfigItem("set_title", True),
+            ConfigItem("pipe_direct_mpv", False),
             ConfigItem("mpris", not mswin),
             ConfigItem("show_qrcode", False),
             ConfigItem("history", True), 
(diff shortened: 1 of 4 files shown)


barracuda156:
@thatonearchguy Is it possible to support custom players? I would like to use `QMPlay2` for video output, and have the same issue with only low-res video playing.

eligiobz:
@thatonearchguy Hi, I'm trying this setting the playerargs like so. 

```
--ytdl-format="bestvideo[height<=?2160][fps<=?60][vcodec!=?vp9]+bestaudio/best"
```

Also tried removing double dash and the quotes and using single quotes (and the combinations) and I'm still getting the lowres video. Am I missing something?

PR #1334, opened 2026-01-01
Fix entry point script.

Hellseher:
`main.py` is in a "mps_youtube" and not in "mps_youtube/main".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ options = dict(
     url="https://github.com/mps-youtube/yewtube",
     download_url="https://github.com/mps-youtube/yewtube/releases",
     packages=['mps_youtube', 'mps_youtube.commands', 'mps_youtube.listview', 'mps_youtube.players'],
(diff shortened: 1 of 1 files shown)


McSinyx:
To quote the [entry point specification](https://packaging.python.org/en/latest/specifications/entry-points#data-model):

> The object reference points to a Python object. It is either in the form `importable.module`, or `importable.module:object.attr`. Each of the parts delimited by dots and the colon is a valid Python identifier. It is intended to be looked up like this:
> 
> ```python
> import importlib
> modname, qualname_separator, qualname = object_ref.partition(':')
> obj = importlib.import_module(modname)
> if qualname_separator:
>     for attr in qualname.split('.'):
>         obj = getattr(obj, attr)
> ```

In other words, even though some build frontends like pip create entry points with `from importable.module import object; object.attr()`/`from mps_youtube import main; main.main()`, they only need to `from importable import module; module.object.attr`/`import mps_youtube; mps_youtube.main.main` to be compliant with the specification.

PR #1339, opened 2026-02-19
Major refactoring, new features added

al3xxx:
- Eliminated Search error related to proxy (httpx 0.28.1 compatibility)
- Added commands tab completion (for settings and playlists too)
- Added list sorting by name, duration, age
- Applied ruff formatting

I think Gemini broke __init__.py compatibility with pip package, but I hope somebody smarter then me could fix it. I did some manual testing major functionality is there, didn't have chance to test transcoding though will get to it later, I didn't touch that code


azinsharaf:
have you refactored youtube-search-python library sections? i think it is one of the reasons the repo has issues. i think it should be migrated to yt-dlp.

al3xxx:
@azinsharaf Didn't touch youtube-search-python itself, just handling result from it in pafy.py and commands/search.py

al3xxx:
@azinsharaf  Sorry, found a bug with not passing kbd to mpv during playback. Tested fix

al3xxx:
I see that this PR was a bit rush, I'll need to do some more testing, found more bugs

coderamm:
I encountered the bug where keyboard inputs were not passed to mpv during playback. I don't think that's specific to your PR. I'm on Win and switching to mpv.net and updating the default player in YewTube fixed that issue for me (although, it causes other issues - each time I run yt, an empty mpv window opens which needs to be closed)

Issue #1345, opened 2026-03-19
[Error] Wrong config gettings


tienhm0202:
At file: https://github.com/mps-youtube/yewtube/blob/master/mps_youtube/player.py

```python
            try:
                if config.SHOW_VIDEO and config.SHOW_SUBTITLES:
                    self.subtitle_path = pafy.get_subtitles(self.song.ytid, config.DDIR.get)
                self.video, self.stream, self.override = stream_details(
                                                            self.song,
                                                            override=self.override,
                                                            softrepeat=self.softrepeat)
```

This make config is not loaded, should be `if config.SHOW_VIDEO.get and config.SHOW_SUBTITLES.get` like another config getter.

PR #1346, opened 2026-03-19
fix(#1345): player got wrong config getter

tienhm0202:
- SHOW_VIDEO and SHOW_SUBTITLES is object and need `.get` - if not it's always returned True

#1345


JorgeTC (changes requested):
If you allow the condition to be false, you are not ensuring the definition of the attribute `subtitle_path`. It will be accessed if video is reproduced in VLC and will raise an `AttributeError`:
https://github.com/mps-youtube/yewtube/blob/35989588eb1270d55bbced991ea2edf3a11064da/mps_youtube/players/vlc.py#L24

It would be enough to define an empty string attribute if the condition is false.

Sophira:
I'm pretty certain the root problem was already dealt with back in #1326!

[Edit: At least, I thought it was. Did that get reverted?]

[Edit 2: Oh, right, that PR was based on the [`develop`](https://github.com/mps-youtube/yewtube/commits/develop/) branch, since [CONTRIBUTING.md](https://github.com/mps-youtube/yewtube/blob/master/CONTRIBUTING.md) says that all contributions should be made on that branch. There just hasn't been a merge into [`master`](https://github.com/mps-youtube/yewtube/commits/master/) since then.]

[Edit 3: Made #1354 to request a merge.]

